PR #13, opened 2022-04-27
Fix jwk persistence

aspeteRakete:
Fix the jwk parsing from database. Add Tests for the manager.


crypto/jwk/manager.go
```diff
@@ -86,10 +85,15 @@ func (m *DefaultManager) GetSigningKey() (jwk.Key, error) {
 	if err != nil {
 		return nil, err
 	}
-	key, err := jwk.FromRaw(k)
+
+	set, err := jwk.Parse(k)
```
FreddyDevelop, line 89: I would use `jwk.ParseKey` here also, like in `GetPublicKeys` as both functions are using the same dataset, so no need for two different parse functions.

PR #14, opened 2022-04-28
update config

FreddyDevelop:


cmd/root.go
```diff
@@ -25,8 +26,12 @@ func NewRootCmd() *cobra.Command {
 		Use: "hanko",
 	}
 	cmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file")
-	initConfig()
-	err := initPersister()
+	err := initConfig()
+	if err != nil {
+		_, _ = fmt.Fprintln(os.Stderr, err)
```
aspeteRakete, line 31: Should we switch to log.Fatal in the init error cases?

aspeteRakete (approved):

cmd/root.go
FreddyDevelop: Yeah, it might me more meaningful and easier to read than `fmt.Fprintln; os.Exit(1)`.

PR #25, opened 2022-05-09
Feat user admin endpoints

lfleischmann:


handler/user_admin.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (h *UserHandlerAdmin) Delete(c echo.Context) error {
+	userId, err := uuid.FromString(c.Param("id"))
+
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, dto.NewApiError(http.StatusBadRequest))
+	}
+
+	return h.persister.Transaction(func(tx *pop.Connection) error {
```
aspeteRakete, line 28: Do we need a transaction here?

handler/user_admin.go
```diff
@@ -0,0 +1,118 @@
...
+	PerPage int `query:"per_page"`
+	Page    int `query:"page"`
+}
+
+func (h *UserHandlerAdmin) List(c echo.Context) error {
+	var request UserListRequest
+	err := echo.QueryParamsBinder(c).
+		Int("per_page", &request.PerPage).
+		Int("page", &request.Page).
+		BindError()
```
FreddyDevelop, line 106: Why did you do this by hand, when you added struct tags to the type? 

Why not using `(&DefaultBinder{}).BindQueryParams(c, &request)`?

lfleischmann: I don't know. Must have been caught up in the echo docs and examples.

FreddyDevelop (approved):

PR #32, opened 2022-05-12
feat: add docs folder with swagger ui and move api yaml

aspeteRakete:
simple addition to display the whole spec unter https://teamhanko.github.io/hanko


FlxMgdnz (commented):
lgtm

PR #39, opened 2022-05-24
feat: add endpoint which redirects to current user resource

FreddyDevelop:
adds the endpoint `/me` which redirects to `users/:id` when the session is valid.

### DIFF
--- a/docs/spec/api.yaml
+++ b/docs/spec/api.yaml
@@ -394,6 +394,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /me:
+    get:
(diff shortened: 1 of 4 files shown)


docs/spec/api.yaml
```diff
@@ -394,6 +394,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /me:
+    get:
+      summary: 'Redirects to user resource if user has valid session'
+      operationId: IsUSerAuthorized
```
lfleischmann, line 400:
```suggestion
      operationId: IsUserAuthorized
```

lfleischmann (approved):

PR #44, opened 2022-05-25
return userId along with credentialId after webauthn login

FreddyDevelop:


lfleischmann:
Would be nice if you could update the spec too.

docs/spec/api.yaml
```diff
@@ -772,6 +776,16 @@ components:
...
                 type: string
                 format: base64url
+    WebauthnLoginResponse:
+      description: 'Response after a successful login with webauthn'
+      type: object
+      properties:
+        credential_id:
+          type: string
+          format: base64url
+        user_ud:
```
lfleischmann, line 786:
```suggestion
        user_id:
```

PR #67, opened 2022-06-13
get user object only for himself

FreddyDevelop:
Currently every logged in user can get user info for every user from the api (`/users/:id`) when he knows the id of a user. So we have broken authorization for this endpoint. This PR fixes this problem.

### DIFF
--- a/backend/handler/user.go
+++ b/backend/handler/user.go
@@ -58,6 +58,15 @@ func (h *UserHandler) Create(c echo.Context) error {
 func (h *UserHandler) Get(c echo.Context) error {
 	userId := c.Param("id")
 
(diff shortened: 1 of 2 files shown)


backend/handler/user.go
```diff
@@ -58,6 +58,15 @@ func (h *UserHandler) Create(c echo.Context) error {
 func (h *UserHandler) Get(c echo.Context) error {
 	userId := c.Param("id")
 
+	sessionToken, ok := c.Get("session").(jwt.Token)
```
lfleischmann, line 61: Why? It is "protected" by the session middleware, wha do we need this in addition to the middleware?

aspeteRakete (approved):

PR #69, opened 2022-06-13
refactor(hanko-js): rename hanko-ui tag name, the output filename, an…

bjoern-m:
Rename the Hanko element to "hanko-ui" and the output file to "element.hanko-ui.js" to hopefully get closer to best practices.


aspeteRakete:
example/public/html/index.html
should also be updated.

bjoern-m:
> example/public/html/index.html should also be updated.

Many thanks! I just had a discussion about this. I'll rename it again to "hanko-auth" in case we decide to release another web component. Then the name hanko-ui would be confusing.

FreddyDevelop (approved):

PR #79, opened 2022-06-14
fix database column type to store longer webauthn public keys

FreddyDevelop:
Change the database column type for the webauthn public keys, because currently the default length which `fizz` assigns is 255, which is too short for some public keys (like Windows Hello ones). This PR changes the column type to `text`, so that which doesn't have a explicit max length.


aspeteRakete:
LGTM but can't test this.

FlxMgdnz:
Tested this PR with my Windows 10 machine and it worked.

aspeteRakete (approved):

PR #82, opened 2022-06-16
Hanko js readme

bjoern-m:


hanko-js/README.md
```diff
@@ -0,0 +1,81 @@
+# &lt;hanko-auth&gt; element
+
+The `<hanko-auth>` element offers a complete user interface that will bring a modern
+login and registration experience to your users. It integrates the Hanko API,
+a backend that provides the underlying functionalities.
+
+## Features
+
+* Registration and login flows with and without passwords
+* Platform Authenticators (e.g. Apple's Passkeys, Windows Hello, etc.)
```
FlxMgdnz, line 10: I'd prefer "Passkey authentication (on supported platforms)" instead of "Platform Authenticators (e.g. Apple's Passkeys, Windows Hello, etc.)" - Also: let's not say "Apple's passkeys" as it is a general term all platforms are using.

hanko-js/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+```html
+<hanko-auth api="https://hanko.yourdomain.com" lang="en" />
+```
+
+Please take a look at the [Hanko API](../backend/README.md) to see how to spin up the backend.
+
+Note, that we're working on a SaaS solution, so that you don't need to run the
+Hanko API by yourself and everything you need is to add the `<hanko-auth>` element
+to your page.
```
FlxMgdnz, line 47: Note that we're working on Hanko Cloud, so that you don't need to run the
Hanko API by yourself and all you need is to do is adding the `<hanko-auth>` element
to your page.

hanko-js/README.md
```diff
@@ -0,0 +1,81 @@
...
+hanko.addEventListener('success', () => {
+    hanko.parentElement.innerHTML = 'secured content...'
+})
+```
+
+## Demo
+
+The GIF below demonstrates how a user registration with passwords enabled looks like. You can set up the flow you like using the Hanko API
+configuration file. The registration flow also includes email verification via passcodes and the registration of a
+platform authenticator so that the user can log in without passwords or passcodes on the current device.
```
FlxMgdnz, line 72: The animation below demonstrates how user registration with passwords enabled looks like. You can set up the flow you like using the Hanko API configuration file. The registration flow also includes email verification via passcodes and the registration of a passkey so that the user can log in without passwords or passcodes.

hanko-js/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+![](demo.gif)
+
+## Browser support
+
+- Chrome
+- Firefox
+- Safari
+- Microsoft Edge
```
FlxMgdnz, line 81: Should we add versions? Or is this just to say "we tested those browsers and it should work there"?

FlxMgdnz (approved):

hanko-js/README.md
bjoern-m: I changed it, but left out "on supported platforms" because it's clear that it won't work then.

hanko-js/README.md
bjoern-m: Thanks, updated!

hanko-js/README.md
bjoern-m: Thanks, updated!

hanko-js/README.md
bjoern-m: I currently wouldn't know which exact versions are working. Would suggest to create a ticket. We might be able to take that into account for test automation as well, but I am not sure about that right now.

FlxMgdnz (approved):

PR #106, opened 2022-06-30
Chore licences

aspeteRakete:
Relicense only the backend part as AGPL and the frontend part as MIT


FreddyDevelop:
Now, the example and the other stuff that is not in one of the two folders has no license. I think we should orientate on how [cal.com](https://github.com/calcom/cal.com/blob/main/LICENSE) has done it, although its a bit different, because they have an enterprise license.

FreddyDevelop (approved):

FlxMgdnz (approved):

PR #109, opened 2022-06-30
check min & max length of a password at login & set new password

FreddyDevelop:


backend/config/config.go
```diff
@@ -157,7 +160,8 @@ func (s *Service) Validate() error {
 }
 
 type Password struct {
-	Enabled bool
+	Enabled           bool
+	MinPasswordLength int `koanf:"min_password_length"`
```
aspeteRakete, line 164: Please also add yaml annotations.

aspeteRakete, line 164:
```suggestion
	MinPasswordLength int `koanf:"min_password_length" yaml:"min_password_length"`
```

backend/handler/password.go
```diff
@@ -49,6 +56,14 @@ func (h *PasswordHandler) Set(c echo.Context) error {
 		return dto.NewHTTPError(http.StatusBadRequest, "failed to parse userId as uuid").SetInternal(err)
 	}
 
+	pwBytes := []byte(body.Password)
+	if utf8.RuneCountInString(body.Password) < h.cfg.MinPasswordLength { // use utf8.RuneCountInString, so utf8 characters would count as 1
+		return dto.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("password must be at least %d characters long", h.cfg.MinPasswordLength))
+	}
+	if len(pwBytes) > 72 {
```
aspeteRakete, line 63: https://cs.opensource.google/go/x/crypto/+/master:bcrypt/bcrypt_test.go;l=67
The actual max Length of this implementation seems to be 55

backend/handler/password.go
```diff
@@ -49,6 +56,14 @@ func (h *PasswordHandler) Set(c echo.Context) error {
 		return dto.NewHTTPError(http.StatusBadRequest, "failed to parse userId as uuid").SetInternal(err)
 	}
 
+	pwBytes := []byte(body.Password)
+	if utf8.RuneCountInString(body.Password) < h.cfg.MinPasswordLength { // use utf8.RuneCountInString, so utf8 characters would count as 1
```
aspeteRakete, line 60: Why check for RuneCount? I thought the bcrypt algorithm operates on bytes, so if you use utf-8 characters you could potentially break the password length limit.

FreddyDevelop: Yes, thats true, but if you want to show the user, whats wrong, you don't want to bother the user with "bytes", because most of the user don't know what the term "bytes" means and don't understand what they should do to avoid that error. So I think it's better to check for characters and not for bytes here.

backend/handler/password.go
FreddyDevelop: Hmm interressting. I wrote some example test and it works with 72 bytes.

```
func TestName(t *testing.T) {
	hash, err := bcrypt.GenerateFromPassword([]byte("012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"), 12)
	assert.NoError(t, err)
	println(string(hash))

	err = bcrypt.CompareHashAndPassword(hash, []byte("012345678901234567890123456789012345678901234567890123456789012345678901"))
	assert.NoError(t, err)
}
```

So the first password has more than 72 bytes and it cuts off the everything greater than 72 bytes, because the second password has exactly 72 bytes. If you remove the last byte/character (the "1") then the test fails. So I think the limit is 72 bytes. See also https://github.com/golang/go/issues/36546

aspeteRakete: Alright! Could you add that test to the suite for documentation purposes? 👍

aspeteRakete (approved):

PR #131, opened 2022-07-13
Feat handle allowed credentials and transports

bjoern-m:
* Improves login flow: enter a known email, click "Continue" and the WebAuthN dialog will be displayed if a credential was previously registered.
* Credential transports will be stored
* Allowed credentials will be transmitted during authentication

TODO: Update e2e tests

### DIFF
--- a/backend/persistence/webauthn_credential_persister.go
+++ b/backend/persistence/webauthn_credential_persister.go
@@ -47,6 +47,17 @@ func (p *webauthnCredentialPersister) Create(credential models.WebauthnCredentia
 		return fmt.Errorf("credential object validation failed: %w", vErr)
 	}
 
+	// Eager creation seems to be broken, so we need to store the transports separately.
+	// See: https://github.com/gobuffalo/pop/issues/608
+	vErr, err = p.db.ValidateAndCreate(&credential.Transports)
+	if err != nil {
+		return fmt.Errorf("failed to store credential transport: %w", err)
+	}
+
+	if vErr != nil && vErr.HasAny() {
+		return fmt.Errorf("credential transport validation failed: %w", vErr)
+	}
+
 	return nil
 }
 
@@ -74,7 +85,7 @@ func (p *webauthnCredentialPersister) Delete(credential models.WebauthnCredentia
 
 func (p *webauthnCredentialPersister) GetFromUser(userId uuid.UUID) ([]models.WebauthnCredential, error) {
 	var credentials []models.WebauthnCredential
-	err := p.db.Where("user_id = ?", &userId).All(&credentials)
+	err := p.db.Eager().Where("user_id = ?", &userId).All(&credentials)
 	if err != nil && err == sql.ErrNoRows {
 		return nil, nil
 	}
--- a/backend/persistence/webauthn_credential_persister.go
+++ b/backend/persistence/webauthn_credential_persister.go
@@ -37,6 +37,8 @@ func (p *webauthnCredentialPersister) Get(id string) (*models.WebauthnCredential
 	return &credential, nil
 }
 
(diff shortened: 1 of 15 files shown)


hanko-js/src/lib/HankoClient.ts
```diff
@@ -295,6 +297,8 @@ class WebauthnClient extends AbstractClient {
           throw new WebAuthnRequestCancelledError(e);
         })
         .then((attestation: PublicKeyCredentialWithAttestationJSON) => {
+          // @ts-ignore
+          attestation.transports = attestation.response.transports;
```
bjoern-m, line 301: @FreddyDevelop do you have any clue what we can do here?

FreddyDevelop: Well, you can leave it like it is and add a comment why this is needed.  
You can also create a new type and use it to send the attestation.

Both ways are acceptable for me.

bjoern-m: Hmm, what should the comment say and what is reason the types do not match? Or does it behave as expected?

bjoern-m: Alright, thanks a lot! I did both, added a new interface and a comment.

backend/persistence/webauthn_credential_persister.go
```diff
@@ -47,6 +47,17 @@ func (p *webauthnCredentialPersister) Create(credential models.WebauthnCredentia
...
+	// See: https://github.com/gobuffalo/pop/issues/608
+	vErr, err = p.db.ValidateAndCreate(&credential.Transports)
+	if err != nil {
+		return fmt.Errorf("failed to store credential transport: %w", err)
+	}
+
+	if vErr != nil && vErr.HasAny() {
+		return fmt.Errorf("credential transport validation failed: %w", vErr)
+	}
+
```
FreddyDevelop, line 60: I don't know if i like this, because what happens when inserting the transports fails, then the credential itself will be stored in the DB, but not the transports and an error will be returned and most likely the endpoint which called this function will retune an error too. So the outcome of this is, that we have an (maybe) unused credential in the DB without any transports, although an error was returned.

I know, that this function is currently called only once and it is wrapped in a database transaction, but if we use this function elsewhere in the future we might have forgotten this behaviour.

I think we should either separate storing the credential and its transports in two different functions or at least add a comment to this function that if should only be called in a database transaction.

bjoern-m: I don't like it myself. Had to do it this way because eager creation doesn't work. The point about needing a transaction is of course true, but that doesn't change if I create two different methods. Also currently a transaction is used. I thought about splitting the method in two parts, but where to put the new method? Would I create a new persister, because one persister takes care of one entity type?

hanko-js/src/lib/HankoClient.ts
```diff
@@ -317,13 +327,13 @@ class WebauthnClient extends AbstractClient {
     });
   }
 
-  isSupported() {
+  isAuthenticatorSupported() {
```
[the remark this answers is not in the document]
bjoern-m: The web component, or any other UI built on top of the HankoClient, must also be able to perform this check. And so it always happens on the same basis, the client package provides this functionality.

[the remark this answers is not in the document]
FreddyDevelop: 👍

backend/persistence/webauthn_credential_persister.go
FreddyDevelop: After thinking more about that, we should leave it like it is (for now) but add a comment to the `Create()` function, that it should only be called inside a database transaction.

FreddyDevelop (approved):

PR #135, opened 2022-07-18
feat: handle auth-token header and create cross domain cookies

bjoern-m:
* Added a new config parameter `EnableAuthToken` to the session config. This causes the API to issue a JWT via the `X-Auth-Token` header. 
* The HankoClient (.ts) will notice and create a client-side cookie containing the value of `X-Auth-Token`

Sidenotes:
* In the different handlers, due to the introduction of the new config parameter "EnableAuthToken", I needed to access the session config. Since all configs are passed separately as parameters into the handler constructors, I would have had to pass a third config in some cases. But all configs are members of the `config.Config{}` type. So instead of creating another parameter, I changed the handlers to accept `config.Config{}`. This way there is no need to update the router and the test files again, when a handler needs to access another config.

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -303,8 +303,9 @@ func (s *Secrets) Validate() error {
 }
 
 type Session struct {
-	Lifespan string `yaml:"lifespan" json:"lifespan" koanf:"lifespan"`
-	Cookie   Cookie `yaml:"cookie" json:"cookie" koanf:"cookie"`
+	EnableAuthToken bool   `yaml:"enable_auth_token" json:"enable_auth_token" koanf:"enable_auth_token"`
+	Lifespan        string `yaml:"lifespan" json:"lifespan" koanf:"lifespan"`
+	Cookie          Cookie `yaml:"cookie" json:"cookie" koanf:"cookie"`
 }
 
 func (s *Session) Validate() error {
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -303,9 +303,9 @@ func (s *Secrets) Validate() error {
 }
 
 type Session struct {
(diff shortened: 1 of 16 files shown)


hanko-js/src/lib/HankoClient.ts
```diff
@@ -96,20 +102,32 @@ class HttpClient {
...
+      const cookieName = "hanko";
+      const token = Cookies.get(cookieName);
 
       fetch(this.api + path, {
         mode: "cors",
         credentials: "include",
         headers: {
           Accept: "application/json",
           "Content-Type": "application/json",
+          Authorization: `Bearer ${token}`,
```
FreddyDevelop, line 114: This is not optimal, because if the cookie can not be read, the header looks like this `Authorization: Bearer undefined`. This feels a little bit buggy. We should not send this header, if the cookie is `undefined `.

backend/config/config.go
```diff
@@ -303,8 +303,9 @@ func (s *Secrets) Validate() error {
 }
 
 type Session struct {
-	Lifespan string `yaml:"lifespan" json:"lifespan" koanf:"lifespan"`
-	Cookie   Cookie `yaml:"cookie" json:"cookie" koanf:"cookie"`
+	EnableAuthToken bool   `yaml:"enable_auth_token" json:"enable_auth_token" koanf:"enable_auth_token"`
```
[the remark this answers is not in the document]
bjoern-m: Makes sense, renamed to `EnableAuthTokenHeader`

hanko-js/src/lib/HankoClient.ts
bjoern-m: Fixed this one. I also rebased with main and did the same thing to the `_fetch2` function (Safari/iOS workaround).

FreddyDevelop (approved):

PR #154, opened 2022-07-26
fix(hanko-js): fix issue with user activated events on iOS 15

bjoern-m:


hanko-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@teamhanko/hanko-elements",
-  "version": "0.0.3-alpha",
+  "version": "0.0.6-alpha",
```
lfleischmann, line 3: Can you update the lockfile accordingly? I guess its main concern is third party dependency versions but it also keeps track of the root package version and it just doesn't feel right when they're out of tune.

hanko-js/src/lib/HankoClient.ts
```diff
@@ -131,6 +179,11 @@ class HttpClient {
     });
   }
 
+  // _post2 part of a workaround for a Safari bug. See _fetch2().
```
lfleischmann, line 182: Doc comment indicates the name of the function is `_post2` but it is `post2`.

```suggestion
  // post2 part of a workaround for a Safari bug. See _fetch2().
```

or rename the function to `_post2`.

lfleischmann (approved):

PR #155, opened 2022-07-27
fix(hanko-js): fixed sign in with email+webauthn on safari/ios

bjoern-m:


FreddyDevelop:
It does not work the same as before: When you register a user A and an credential on device A and want to login with the user A on device B, then webauthn is triggered (which is the same behaviour as before) and because you have no credential on device B, you cannot sign the challenge, so you cancel the webauthn browser dialog and now you have no way of sign in to the account, because passcode is not triggered.

bjoern-m:
> It does not work the same as before: When you register a user A and an credential on device A and want to login with the user A on device B, then webauthn is triggered (which is the same behaviour as before) and because you have no credential on device B, you cannot sign the challenge, so you cancel the webauthn browser dialog and now you have no way of sign in to the account, because passcode is not triggered.

Nice, thanks! I fixed that one. @FreddyDevelop

FreddyDevelop (approved):
LGTM

PR #159, opened 2022-08-01
build & publish hanko-js docker container

FreddyDevelop:
Build & publish the hanko-js docker container.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -33,6 +33,8 @@ jobs:
             image: ghcr.io/teamhanko/hanko
           - context: ./example
             image: ghcr.io/teamhanko/hanko/example
+          - context: ./hanko-js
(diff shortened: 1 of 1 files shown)


lfleischmann (commented):
LGTM, Actions ran successfully but I wondered: [This](https://github.com/orgs/teamhanko/packages?repo_name=hanko) and the `Packages` item on the right-hand sidebar on the main repo page still only list `hanko` and `hanko-example` packages. Is `hanko-js` supposed to appear there as well and if so, is there any additional configuration required? If not: why shouldn't it appear there?

FreddyDevelop:
This is because the publish step is only done when its not in a PR. See [here](https://github.com/teamhanko/hanko/blob/a5e55792ccbd3b84a22b05061b7dc3c62840ef56/.github/workflows/docker-publish.yml#L83)

lfleischmann (approved):

PR #163, opened 2022-08-02
Found Broken Link under Licenses

SojinSamuel:
Under licenses, internal link AGPL-3.0 returned a 404 error. the broken internal link was updated.


FlxMgdnz:
Good catch, thanks!

PR #166, opened 2022-08-03
fix: remove webauthn transports when on android

FreddyDevelop:
This PR removes the transports from the `allowCredentials` when webauthn is request on Android.

When the transports are not removed and only `internal` is set at transports, then Android does not check if it knows one of the credentials in the list and shows the BiometricPrompt and after the biometric were provided, the Web Component switches to Passcode, because after the biometrics are provided, an generic error (`DOMException: The operation either timed out or was not allowed`) is returned. When the transports are not available, then a chooser is shown where the user can select the transport the authenticator will use, but the `internal` authenticator is not offered. 

Prerequisites: The android device has no stored credential for the requested user.

With this change, the UI does not seem buggy after the biometrics were provided. But it is not great, because the user must cancel the transport chooser before passcodes can be used.

### DIFF
--- a/hanko-js/src/lib/HankoClient.ts
+++ b/hanko-js/src/lib/HankoClient.ts
@@ -330,17 +330,10 @@ class WebauthnClient extends AbstractClient {
           reject(e);
         })
         .then((challenge: CredentialRequestOptionsJSON) => {
-          if (
-            challenge.publicKey.allowCredentials != undefined &&
-            this.isAndroidUserAgent
-          ) {
-            for (
-              let i = 0;
-              i < challenge.publicKey.allowCredentials.length;
-              i++
-            ) {
-              delete challenge.publicKey.allowCredentials[i].transports;
-            }
+          if (this.isAndroidUserAgent) {
+            challenge.publicKey?.allowCredentials?.map((c) => {
+              delete c.transports;
+            });
           }
           return getWebauthnCredential(challenge);
         })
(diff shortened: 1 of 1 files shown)


hanko-js/src/lib/HankoClient.ts
```diff
@@ -327,6 +330,18 @@ class WebauthnClient extends AbstractClient {
...
+            this.isAndroidUserAgent
+          ) {
+            for (
+              let i = 0;
+              i < challenge.publicKey.allowCredentials.length;
+              i++
+            ) {
+              delete challenge.publicKey.allowCredentials[i].transports;
+            }
+          }
```
bjoern-m, line 344: Is it possible to simplify the loop? E.g.:
```suggestion
          if (this.isAndroidUserAgent) {
            challenge.publicKey?.allowCredentials?.map((c) => {
              delete c.transports;
            });
          }
```
```

File: hanko-js/src/lib/HankoClient.ts [hunk 2]
```diff
@@ -327,6 +330,18 @@ class WebauthnClient extends AbstractClient {
           reject(e);
         })
         .then((challenge: CredentialRequestOptionsJSON) => {
+          if (
```
bjoern-m, line 333: Would be nice to have a comment here why we need to do this.

bjoern-m (commented):
I can't really try it without an android phone (let me know if there is a way). Therefore only a small code-wise review from my side.

hanko-js/src/lib/HankoClient.ts
FreddyDevelop: Thanks, I added a comment.

bjoern-m (approved):
As I said, I wasn't able to try it out, but the code is approved from my side.

FreddyDevelop:
Closing in favour of https://github.com/teamhanko/hanko/pull/167

PR #171, opened 2022-08-08
hanko-js: export client lib

bjoern-m:
* package path has changed from `@teamhanko/hanko-elements` to `@teamhanko/hanko`
* import the web component with `import "@teamhanko/hanko/hanko-auth"`
* import client with `import { Client, User, ConflictError,... } from "@teamhanko/hanko/client"`
* webpack creates `element.hanko-auth.js` but as a UMD module, containing the web component and the client
* webpack creates another file `hanko-client.js`, containing the client


FreddyDevelop:
The new `hanko-client.js` file should be included in the hanko-js `Dockerfile`, so we can use the client later in the example app and it will work in the docker compose quickstart.

PR #174, opened 2022-08-09
Attempt at #126

### LINKED ISSUE #126: Backend configuration through environment variables not possible for multi-word, underline delimited keys
The backend configuration does not recognize environment variables where a key consists of multiple "words" delimited by an underscore.

## Expected Behavior
When running the application I expect environment variables to be applied even if a key contains multiple "words" delimited by an underscore, i.e. when running `WEBAUTHN_RELYING_PARTY=http:localhost:8080 ./hanko serve public` I expect the given value to be applied to the config parameter that would be defined in the YAML configuration as `webauthn.relying_party`.

## Current Behavior
The value is not applied because when [loading](https://github.com/teamhanko/hanko/blob/main/backend/config/config.go#L40) the config any underscore is replaced by a dot, hence demarcating even underscores in a single key as equivalent to a level of nesting.

ryanhaticus:
Hi there, Team Hanko!

**Context:**
For context, before you read the code, I am by no means a GO developer. In fact, today is the first day I have touched the language at all. The initiative here is awesome and I wanted to try my hand at contributing to what seems to be an already awesome product.

**Intention:**
In this fix, I am attempting to resolve #126 by iterating over the environment variables and replacing `.` characters with `_` characters from back to front until there isn't an error or until an index of `.` no longer exists.

**Conclusion:**
I have no idea of this will work, nor am I sure if the code I wrote even compiles - but nonetheless I'm throwing my wrench into a world of hammers.

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -48,7 +48,7 @@ func Load(cfgFile *string) (*Config, error) {
 			return r
 		}), nil)
 
-		if err != nil || lastAccessor == -1 {
+		if err == nil || lastAccessor == -1 {
 			break;
 		}
 	}
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -38,7 +38,7 @@ func Load(cfgFile *string) (*Config, error) {
 
 	lastAccessor := -1
 
-	for err != nil {
+	for {
 		err = k.Load(env.Provider("", ".", func(s string) string {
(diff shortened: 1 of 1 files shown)


FreddyDevelop:
Thank you for your contribution. Unfortunately this solution does not work. The problem with it is, that koanf does not return an error if keys can't be mapped. So the `for` loop will always only run once.

aspeteRakete:
Closing due to inactivity.

PR #176, opened 2022-08-10
Found explicit type declarations

LucidSamuel:


aspeteRakete:
The file doesn't exist anymore. Fixed in frontend-sdk. Thanks!

PR #214, opened 2022-09-06
Request(roadmap): Push authentication

SimoMay:
One of my favorite passwordless method is push authentication, I'd love to suggest adding it to the Roadmap of Hanko.

Here is a good article about how it works: https://www.twilio.com/blog/understanding-push-authentication

When signing in we can check if the account has other "trusted devices" linked. We either let the user decide if rather wants to auth at the current device or do it through the linked device, or we can force the user to use the linked device, for example, if the current device doesn't support Webauthn and linked device does.


SimoMay:
Closed- Reason: Cross-device signing with passkeys is becoming part of WebAuthn infrastructure and that would be superior to building a proprietary phone / push sign in infrastructure in Hanko

PR #247, opened 2022-09-19
fix: send 401 when passcode is not found instead of 404

### LINKED ISSUE #240: Return same HTTP Code for both passcode ID not found and passcode is incorrect
Currently, the backend returns an HTTP 404 NotFound if the [passcode ID supplied in the request cannot be found](https://github.com/teamhanko/hanko/blob/main/backend/handler/passcode.go#L164) and returns an HTTP 401 Unauthorized response if the [passcode ID is correct but the code supplied is not](https://github.com/teamhanko/hanko/blob/main/backend/handler/passcode.go#L192). Most standard login forms return the same HTTP code (401) if either username or password are incorrect, and since passcodes essentially act as a username/password combination I suggest we return the same HTTP code for either scenario.

Thinking from an adversary's standpoint, letting them know whether a passcode ID exists helps them when brute forcing a bunch of passcode ID and passcode options.

aspeteRakete:
closes #240

### DIFF
--- a/backend/handler/passcode.go
+++ b/backend/handler/passcode.go
@@ -161,7 +161,7 @@ func (h *PasscodeHandler) Finish(c echo.Context) error {
 			return fmt.Errorf("failed to get passcode: %w", err)
 		}
 		if passcode == nil {
-			businessError = dto.NewHTTPError(http.StatusNotFound, "passcode not found")
+			businessError = dto.NewHTTPError(http.StatusUnauthorized).SetInternal(errors.New("passcode invalid"))
 			return nil
 		}
 
--- a/backend/handler/passcode_test.go
+++ b/backend/handler/passcode_test.go
@@ -173,7 +173,7 @@ func TestPasscodeHandler_Finish_WrongId(t *testing.T) {
 	err = passcodeHandler.Finish(c)
 	if assert.Error(t, err) {
 		httpError := dto.ToHttpError(err)
-		assert.Equal(t, http.StatusNotFound, httpError.Code)
+		assert.Equal(t, http.StatusUnauthorized, httpError.Code)
 	}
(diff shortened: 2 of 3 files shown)


aspeteRakete:
The frontend reacts here: https://github.com/teamhanko/hanko/blob/5693c90077a2ea4e332f76a679afecab32d6aa81/frontend-sdk/src/lib/client/PasscodeClient.ts#L99
I don't see a big problem going forward with this.

lfleischmann:
> The frontend reacts here:
> 
> https://github.com/teamhanko/hanko/blob/5693c90077a2ea4e332f76a679afecab32d6aa81/frontend-sdk/src/lib/client/PasscodeClient.ts#L99
> 
> 
> I don't see a big problem going forward with this.

If the handler no longer returns a 404, do we still need to explicitly check for it here:

https://github.com/teamhanko/hanko/blob/5693c90077a2ea4e332f76a679afecab32d6aa81/frontend-sdk/src/lib/client/PasscodeClient.ts#L101

?

aspeteRakete:
> > The frontend reacts here:
> > https://github.com/teamhanko/hanko/blob/5693c90077a2ea4e332f76a679afecab32d6aa81/frontend-sdk/src/lib/client/PasscodeClient.ts#L99
> > 
> > I don't see a big problem going forward with this.
> 
> If the handler no longer returns a 404, do we still need to explicitly check for it here:
> 
> https://github.com/teamhanko/hanko/blob/5693c90077a2ea4e332f76a679afecab32d6aa81/frontend-sdk/src/lib/client/PasscodeClient.ts#L101
> 
> ?

You are right, of course. Fixed.

lfleischmann (approved):

PR #250, opened 2022-09-21
Remove console log

irby:
Just a little cleanup.

Re-doing this PR because of git signing issues.


aspeteRakete:
Thanks Matthew!

PR #251, opened 2022-09-21
fix(hanko-frontend-sdk): access X-Auth-Token header during cross-doma…

bjoern-m:
access X-Auth-Token header during cross-domain in operations only


aspeteRakete:
As discussed I would favor an approach where EnableAuthTokenHeader stays as a config parameter and the cors config gets set accordingly.
Otherwise you have to document a "magic string" to use hanko in a cross-domain context instead of a simple bool setting.

bjoern-m:
Ok I can do that, but just for understanding: What is the purpose of `server.public.cors.expose_headers`? Why would I expose another header?

aspeteRakete:
Talked with @FreddyDevelop about it. We could not find a use case for all the cors settings. The only things we really need is allowOrigins, allowCredentials and maybe maxAge... @lfleischmann do any use cases come to your mind?

bjoern-m:
Okay, thanks for your answers. I now just followed @aspeteRakete 's suggestion and re-introduced the `session.EnableAuthTokenHeader` config parameter and didn't touched the CORS config at all.

aspeteRakete (approved):

PR #255, opened 2022-09-26
Feat conditional UI

bjoern-m:


FreddyDevelop:
I tested it on MacOS Ventura Beta and there is a small problem with conditional UI when you come to the page for the first time (or always when you visit the page, but not at a reload). For example you are on google.com and then go to a website which included the web component, the conditional UI is only triggered when the email input is focused a third time (see video). I think that is because the webauthn request must be created before the input gets focused. 

So i think the request must be created after page load and not after the input element gets focused.

https://user-images.githubusercontent.com/29370474/192816828-281d869d-b563-4647-ac21-16f93751e62a.mov

FreddyDevelop (approved):

PR #315, opened 2022-10-20
refactor: rename example to quickstart

lfleischmann:
# Description

Rename `example` to `quickstart`.


lfleischmann:
> The image in the footer still shows "Example app". Should this be changed to "Quickstart app", so it will not confuse anybody? Screenshot 2022-10-20 at 17 25 57

My thoughts were: the quickstart is still sort of an example. The main benefit of renaming the app/folder/image is for development purposes (easier to distinguish than between `example` and `examples` folder). My impression was that "end-users" of a hosted version (like wo do on `example.hanko.io` or `example.stg.hanko.io`) of the quickstart do not really care whether this is the "quickstart" app or an "example" app. If we were to change this, then we could also ask whether the domain of the hosted examples should be  `quickstart.hanko.io` or `quicksart.stg.hanko.io`. I just didn't think this was necessary.

aspeteRakete (approved):

FreddyDevelop (approved):

PR #331, opened 2022-10-25
feat: remove platform attachment requirement on registration

lfleischmann:
# Description

Remove platform attachment requirement on registration and change `isUserVerifyingPlatformAuthenticatorAvailable` calls to the general Webauthn availability check.

# Implementation

Safari currently (v. 16.2) does not provide a UI in case of a (registration) ceremony being performed with an authenticator NOT protected by e.g. a PIN. While Chromium based browsers do offer a UI guiding through the setup of a PIN, Safari simply performs the ceremony without then setting the UV flag even if it is required. On finalization in the backend, this ultimately leads to an error of in the `webauthn` library which is then "swallowed" by being mapped to a generic error response ("Failed to validate attestation"). In order to provide an appropriate error message to the frontend/user, I had to return an error response distinguishable from other error cases so the `webauthn` handler now returns an HTTP error with a 
dedicated/separate HTTP status code (i.e. NOT "Bad request") because it seemed a bit more robust than forcing the frontend to check on a matching (sub-)string in the error message in order to properly display the error.

# Tests

The changes add e2e test runs to check whether WebAuthn credentials now allow for transports other than `platform` . However, the e2e tests do not test the above mentioned implemented error handling for user verification failures because:

1. The e2e tests only use Chromium, the error handling is more or less a Safari workaround
2. While authenticators used in the virtual WebAuthn environment can be configured to either not have user verification capabilities at all or to simply return a false user verification flag, the WebAuthn API/browser responds with an exception/error, so the authenticator response does not even reach the backend and the above mentioned error handling implementation does not take effect.

## How to test

Run the tests or/and run the [quickstart](https://github.com/teamhanko/hanko/tree/main/quickstart#run-the-quickstart).


FlxMgdnz (approved):

FreddyDevelop (approved):
LGTM

PR #362, opened 2022-11-08
Feat bundle frontend sdk

aspeteRakete:
# Description
hanko elements now bundles the frontend-sdk so we can publish elements and sdk independently.
Only one Dockerfile for frontend-sdk and elements is used and the docker-compose file reflects that.

# Implementation

# Tests

# Todos

# Additional context


frontend/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+
+# elements
+WORKDIR /app/elements
+COPY elements/package.json ./
+COPY elements/package-lock.json ./
+
+RUN npm ci
+COPY ./elements ./
+RUN npm run build
```
FreddyDevelop, line 23: Align the two building steps:
```suggestion
# frontend-sdk
WORKDIR /app/frontend-sdk
COPY frontend-sdk/package.json ./
COPY frontend-sdk/package-lock.json ./

RUN npm ci --silent
COPY ./frontend-sdk ./
RUN npm run build


# elements
WORKDIR /app/elements
COPY elements/package.json ./
COPY elements/package-lock.json ./

RUN npm ci --silent
COPY ./elements ./
RUN npm run build
```

frontend/Dockerfile.debug
```diff
@@ -0,0 +1,31 @@
...
+
+
+# elements
+WORKDIR /app/elements
+COPY elements/package.json ./
+COPY elements/package-lock.json ./
+
+RUN npm ci
+COPY ./elements ./
+RUN npm run build:dev
```
FreddyDevelop, line 23: Would be nice if the steps here would also align, like the one in the normal `Dockerfile`, except for the `build:dev` part.

.github/workflows/docker-publish.yml
```diff
@@ -33,9 +33,9 @@ jobs:
             image: ghcr.io/teamhanko/hanko
           - context: ./quickstart
             image: ghcr.io/teamhanko/hanko/quickstart
-          - context: ./elements
+          - context: ./frontend
```
FreddyDevelop, line 36: Is it correct that here the bundled docker image with `elements` and `frontend-sdk` should be build and published and not the one which only contains the `element`?

If so, why do we still have a different image for `frontend-sdk`?

FreddyDevelop:
Dependabot should also be updated to reflect the new folder structure.

frontend/Dockerfile.debug
aspeteRakete: 👍

.github/workflows/docker-publish.yml
aspeteRakete: Good question. We now have the elements docker image that includes the frontend-sdk. Would be sufficient for us to use just this image. Idk if anyone will use the frontend-sdk only. I would leave it for now as it doesn't do any harm and we could take an look at the donwload numbers in the future to determine if it's needed.

FreddyDevelop: 👍

FreddyDevelop (approved):

FreddyDevelop (approved):

PR #393, opened 2022-11-17
Add total count & link header for pagination

FreddyDevelop:
# Description

Adds the `Link` and `X-Total-Count` header to list audit logs and list users. The `X-Total-Count` header returns the total count of the requested resource. The `Link` header returns urls which can be used for pagination. 

E.g. there are 155 registered users and the user list is requested with `?page=5&per_page=10`, then
- the `X-Total-Count` header returns 155
- the `Link` header returns `<http://localhost:8001/users?page=1&per_page=10>; rel="first",<http://localhost:8001/users?page=16&per_page=10>; rel="last",<http://localhost:8001/users?page=6&per_page=10>; rel="next",<http://localhost:8001/users?page=4&per_page=10>; rel="prev"`

# Tests

To test this just use the admin audit_log and admin users api and look into the returned headers.

# Additional context

https://docs.github.com/en/rest/guides/traversing-with-pagination

### DIFF
--- a/backend/handler/user_admin_test.go
+++ b/backend/handler/user_admin_test.go
@@ -281,6 +281,7 @@ func TestUserHandlerAdmin_List(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, 2, len(users))
 		assert.Equal(t, "2", rec.Header().Get("X-Total-Count"))
+		assert.Equal(t, "<http://example.com/users?page=1&per_page=20>; rel=\"first\"", rec.Header().Get("Link"))
 	}
 }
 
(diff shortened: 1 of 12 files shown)


backend/handler/user_admin_test.go
```diff
@@ -280,6 +280,7 @@ func TestUserHandlerAdmin_List(t *testing.T) {
 		err := json.Unmarshal(rec.Body.Bytes(), &users)
 		assert.NoError(t, err)
 		assert.Equal(t, 2, len(users))
+		assert.Equal(t, "2", rec.Header().Get("X-Total-Count"))
```
lfleischmann, line 283: Why check `X-Total-Count` header only? How about the `Link` header?

backend/pagination/header.go
```diff
@@ -0,0 +1,48 @@
+package pagination
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+func CreateHeader(u *url.URL, total int, page int, itemsPerPage int) string {
```
lfleischmann, line 9: How about a unit test for this one?

lfleischmann (approved):

PR #416, opened 2022-11-28
Multiplatform docker

hilli:
# Description

I am doing development on an Apple Silicon Mac and doing deploys to other ARM based hosts. I have previously build the images myself, but figured that I can't be the only one who would like hanko on more than one architecture.

# Implementation

- Added parameter in the Dockerfile's to enable building for multiple architectures
  - Fixing the builder to the BUILDPLATFORM
  - Utilizing GOs crossplatform builder to target TARGETARCH
  - Let `buildx` handle the matrix (1x2, but not the less :)) from there
- Bumped Docker's Github Actions versions to get rid of warnings, hopefully
- Bumped the go version to 1.19 in the Dockerfile

# Tests

Try running the action related to the workflow on this branch. There should be bulds for both amd64 and arm64 on ghcr.io afterwards.

# Todos

There are other architectures available (linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6), which could be added to the matrix, but I would wait with that until there is a need for it.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -58,7 +58,7 @@ jobs:
       # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v1.2
+        uses: docker/login-action@v2
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -48,15 +48,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      # Workaround: https://github.com/docker/build-push-action/issues/461
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Setup Docker buildx
-        uses: docker/setup-buildx-action@v1.6.0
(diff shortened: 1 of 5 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -48,15 +48,17 @@ jobs:
...
+
       - name: Setup Docker buildx
-        uses: docker/setup-buildx-action@v1.6.0
+        uses: docker/setup-buildx-action@v2
 
       # Login against a Docker registry except on PR
       # https://github.com/docker/login-action
       - name: Log into registry ${{ env.REGISTRY }}
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v1.14.1
```
like-a-bause, line 59: Did you want to update to v2 ?

.github/workflows/docker-publish.yml
```diff
@@ -48,15 +48,17 @@ jobs:
...
+        uses: docker/login-action@v1.2
```
like-a-bause, line 61: I wanted to mark this line. :)

Did you want to update to v2 ?

.github/workflows/docker-publish.yml
hilli: Same here.

hilli: Cause it's the same place ;)

like-a-bause (approved):
Awesome! Thanks a lot!

Issue #213, opened 2022-09-06
6 digits passcode can be brute forced 


adri:
Hi! Great work so far. I saw that you use a 6 digit passcode which has just 1 million possibilities. https://github.com/teamhanko/hanko/blob/main/backend/crypto/passcode.go

Without a rate limit this can be brute forced by trying out all combinations. 
I saw that rate limits have to be added by the user, however most IP based rate limits can be bypassed using proxy services. Using only a rate limit is only one layer of defense.

A more secure solution would be to use letters as well so the code is not so easily guessable. That way there is some “defense in depth”.

irby:
I agree passcodes can be brute-forced. The backend passcode handler does invalidate (and erase) passcodes [after 3 incorrect guesses](https://github.com/teamhanko/hanko/blob/main/backend/handler/passcode.go#L178), but I don't think rate limiting is a bad idea to prevent malicious users spamming the API until they get a right passcode ID and token combination.

FlxMgdnz:
We'll address this here (with https://github.com/teamhanko/hanko/issues/24)

PR #501, opened 2023-01-23
Feat basic rate limiting

### LINKED ISSUE #24: Rate limiting (application layer)
Currently we don't have a backoff strategy for passwords and passcodes, but we should implement one.

### Proposal:

**Passwords:**
After 3 "free" tries the backoff takes over. (exponential or linear should be discussed)
Why "free" tries, because sometimes a user types in the wrong password, but immediately remembers and types in the correct one, when the user must wait then, he will be annoyed.

**Passcodes:**
Each passcode has 3 tries to be entered correctly. After that a new passcode must be requested. 
Requesting the second (third?) passcodes will trigger the backoff strategy.

like-a-bause:
# Description
Implements basic fixed window rate limiting.

Fixes #24 Relates to #213 
# Implementation
 Uses https://github.com/sethvargo/go-limiter
Supports both in-memory and redis backend for scalability.
Now only implemented for: password/init and password/login, with separate stores.
Uses a key combination of user-id and ip.
Limits can be defined for each endpoint separateley.

# Tests

Added some tests for the config, which should be expanded for Fields that get Validated.
Added basic functionality test for the rate-limiter library.

# Todos

# Additional context


FreddyDevelop (changes requested):
The different limits for passcode and password do not work when redis is enabled.

FreddyDevelop (approved):
LGTM

FreddyDevelop (approved):

PR #522, opened 2023-01-26
feat: add query params to search users

FreddyDevelop:
# Description

Adds query params to the list users endpoint in the admin API.

# Tests

Try to get users through the admin API and use the query parameters to filter the results.


bjoern-m (approved):
LGTM

PR #534, opened 2023-01-27
chore(hanko-elements): add a new shadow part 'form-item'

bjoern-m:
# Description

Adds a new shadow part to style the form items. One use-case would be that you may not want to have the input field and the button next to each other, even when there would be enough space on the screen:

```
hanko-auth::part(form-item) {
  min-width: 100%; /* input fields and buttons are on top of each other  */
}
```


FreddyDevelop (approved):
It works, but we should document that this and `--input-min-width` might interfere with each other. 

**Example:**
```
hanko-auth::part(form-item) {
    max-width: 50%;
}
```
Screenshot 2023-01-30 at 09 26 11

Issue #550, opened 2023-02-03
Provide a `logout` method on the SDK


IgnisDa:
### Checklist

- [X] I could not find a solution in the existing issues or docs.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/teamhanko/hanko/blob/main/CODE_OF_CONDUCT.md).

### Description

It would be great if the Hanko SDK provided a `logout` method that would remove the `hanko_cookie` from the frontend.

### Describe your ideal solution

Ideally this can be a part of the [`UserClient`](https://docs.hanko.io/jsdoc/hanko-frontend-sdk/UserClient.html).

### Hanko Version

v0.3.2

irby:
I wrote a similar logout method for this project when I forked this project a while ago. I can throw together something that should get logout functionality available from the SDK.

IgnisDa:
@irby Could you please provide the code that does that? I would like to add it manually to my code until you make a PR (or I could make a PR too if you want). Thanks!

irby:
@IgnisDa I'm about 85% done with the PR, and should have it posted here in the next day or so. But, here's the code I wrote some months ago that will expose an endpoint to log out the user.

Updating the Session Manager to have a DeleteCookie method: https://github.com/irby/passwordless-authenticator-poc/blob/main/authenticator-backend/session/session.go#L182

Adding the DeleteCookie method within the Webauthn Test: https://github.com/irby/passwordless-authenticator-poc/blob/main/authenticator-backend/handler/webauthn_test.go#L203

Adding a logout method on user handler: https://github.com/irby/passwordless-authenticator-poc/blob/main/authenticator-backend/handler/user.go#L545

Logout endpoint exposed on the public server: https://github.com/irby/passwordless-authenticator-poc/blob/main/authenticator-backend/server/public_router.go#L69

Once this is added, you should be able to call the logout with POST `/users/logout`. 


While making my PR, I noted some opportunities to improve my code. So, my PR will have some different code than what is in here. Just FYI 😅

bjoern-m:
Hello everyone! 

The logout feature has been merged and the next release will take place next week.

Thanks for contributing!

PR #566, opened 2023-02-09
Feat: Add logout method to sdk

irby:
# Description

Addressing Issue #550 

Adds a log out method to the backend that the SDK can call. The motivation for this is that the Hanko session cookie is a Secure HTTP Only cookie, which cannot be modified by JavaScript for XSS protection reasons. 

# Implementation

Created a new endpoint `/logout` that's exposed on the public API routes. It calls the `UserHandler` new method `Logout` which will call upon the `SessionManager` `DeleteCookie` method. This new method has the same logic as what was in the logout endpoint in `main.go`from the quickstart project. 

In the SDK, I added a new method in `UserClient` called `logout`, which will call the `/logout` API endpoint. If this call returns an unsuccessful status code, it will throw an exception.

# Tests

To verify logic, I added unit tests for the `SessionManager.DeleteCookie` method and the `UserHandler.Logout` method. Added tests for the SDK `UserClient.logout` method that verifies successful API calls do not throw an error, while any bad status code throws the expected error.

# Todos

- ~~The logout button formatting is off. I do not know why the button appears different from the Continue button visible on the login screen.~~
  - I have figured out the resolution for this. I needed to update the common.css file to include the new `hanko-logout` element.
  
  image

- I could not figure out how to update the documentation for the frontend / backend. Are these manually updated or does Docusaurus compile it? I tried running `npm run build` but didn't make any updates.

- What should happen if token expires by the time the logout is called? The HTTP Code for an expired token will be 401. Should we treat this as an OK, logout successful, or should we still throw UnauthorizedError in the SDK?


bjoern-m:
Hi @irby!

Thanks for your effort and for contributing to Hanko! About your questions:

> I could not figure out how to update the documentation for the frontend / backend. Are these manually updated or does Docusaurus compile it? I tried running npm run build but didn't make any updates.

You can run `npm run docs` and find the re-generated files under `/docs/static/jsdoc`. The backend specification can be updated by modifing `docs/static/spec/public.yaml`.

> What should happen if token expires by the time the logout is called? The HTTP Code for an expired token will be 401. Should we treat this as an OK, logout successful, or should we still throw UnauthorizedError in the SDK?

I would say that we should treat it as an OK, because otherwise whoever integrates the functionality has to check the error just to treat is the same way (e.g. redirect to a login page) and any other error, whether it is an internal server error, a bad request, etc. I would map to the "Technical Error" as the user-facing error.

I went through the code and it looked good so far, but I needed to tweak it a little bit to make it work with cross-domain setups, so I created a new branch with some suggestions, fixes, additional functionality and the updated docs. Unfortunately, I can't push it to your repository (permission denied). Do you have any idea how I can share the code with you? Haven't done this before in GitHub. Maybe I missed something, or simply do not have permission?

Regarding the logout button component, we voted in the team not to include it, because we have tentative plans that might interfere and also the components are currently meant to include the error handling. For granular functions like the logout we would recommend to use the SDK and directly call the logout function. Would that be a practical solution for you? And would it be okay to leave out the button-component for now?

Let me know what you think, and thanks again for contributing!

irby:
Hi @bjoern-m I have added you as a contributor to my forked repo. You should be able to push changes to it. Let me know if you run into any issues. 

I am ok with leaving out the button component. I should be able to call the SDK from the quickstart project? 

Once you push your changes, I will:
- Regenerate the docs given the instructions you've provided.
- Remove the logout button component
- Call the SDK directly from the quickstart project
- Change the SDK behavior from throwing an Unauthorized error to an OK? Can you elaborate on "I would map to the "Technical Error" as the user-facing error."?

bjoern-m:
Hi @irby,

it worked and I was able to push my branch and to create a [PR](https://github.com/irby/hanko/pull/1) in your repo! Thanks a lot! 

 >  I should be able to call the SDK from the quickstart project?

Yes, that should be possible.

> Once you push your changes, I will:
>
> * Regenerate the docs given the instructions you've provided.
> * Remove the logout button component
> * Call the SDK directly from the quickstart project

Sounds good! 

> Change the SDK behavior from throwing an Unauthorized error to an OK? Can you elaborate on "I would map to the "Technical Error" as the user-facing error."?

I just wanted to point out that there are many cases where the frontend receives a more specific response code but has no better option than display a message like "Something went wrong" which corresponds to the `TechnicalError`. I've covered this in the PR. The logout either works or it throws the technical error.

irby:
Hi @bjoern-m sorry for the delay on this -- was traveling all last week. 

I'm currently having trouble calling the SDK from the quickstart project. Ideally, I'd like to call the `UserClient.logout` project from within the script section of the `secured.html` file. Do you have any advice on calling this function from this file?

irby:
@bjoern-m 

Thank you for the help on calling the SDK from the quickstart project! I was stuck on that one for a bit. 😆 

I've caught up all the code with what's currently in master and had to fix up a couple of broken things, but everything should be good to go now!


Thanks,
Matt

quickstart/public/assets/css/common.css
```diff
@@ -60,7 +60,7 @@ body {
     padding-bottom: 2rem;
 }
 
-hanko-auth, hanko-profile {
+hanko-auth, hanko-profile, hanko-logout {
```
bjoern-m, line 63: `hanko-logout` can be removed:

```suggestion
hanko-auth, hanko-profile {
```

irby: @bjoern-m done.

bjoern-m:
Hi @irby!

Very nice - looks good to me now. Thanks again for the effort!

bjoern-m (approved):

like-a-bause:
@bjoern-m I think you have to merge this.
@irby Thank you for your contribution!! 🥇

Issue #548, opened 2023-02-01
hanko-elements lang not optional


like-a-bause:
### Checklist

- [X] I could not find a solution in the existing issues or docs.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/teamhanko/hanko/blob/main/CODE_OF_CONDUCT.md).

### Describe the bug

The doc says:
```
lang Currently supported values are "en" for English and "de" for German. If the value is omitted, "en" is used.
```
Tried to update the examples but they wouldn't work with some undefined error.

Providing the lang parameter fixes this. 

### Reproducing the bug

"@teamhanko/hanko-elements": "^0.1.1-alpha"

 <hanko-auth api={api}/>

### Logs

```Shell
elements.js:2 Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
    at t.default [as constructor] (elements.js:2:1)
    at y.L [as render] (elements.js:2:1)
    at I (elements.js:2:1)
    at O (elements.js:2:1)
    at I (elements.js:2:1)
    at O (elements.js:2:1)
    at I (elements.js:2:1)
    at O (elements.js:2:1)
    at I (elements.js:2:1)
    at U (elements.js:2:1)
t.default @ elements.js:2
L @ elements.js:2
I @ elements.js:2
O @ elements.js:2
I @ elements.js:2
O @ elements.js:2
I @ elements.js:2
O @ elements.js:2
I @ elements.js:2
U @ elements.js:2
l @ elements.js:2
a @ elements.js:2
(anonymous) @ elements.js:2
(anonymous) @ elements.js:2
a @ elements.js:2
d @ elements.js:2
(anonymous) @ elements.js:2
(anonymous) @ elements.js:2
a @ elements.js:2
t.register @ elements.js:2
(anonymous) @ HankoAuth.tsx:16
commitHookEffectListMount @ react-dom.development.js:23150
commitPassiveMountOnFiber @ react-dom.development.js:24926
commitPassiveMountEffects_complete @ react-dom.development.js:24891
commitPassiveMountEffects_begin @ react-dom.development.js:24878
commitPassiveMountEffects @ react-dom.development.js:24866
flushPassiveEffectsImpl @ react-dom.development.js:27039
flushPassiveEffects @ react-dom.development.js:26984
(anonymous) @ react-dom.development.js:26769
workLoop @ scheduler.development.js:266
flushWork @ scheduler.development.js:239
performWorkUntilDeadline @ scheduler.development.js:533
```

[...]

PR #569, opened 2023-02-13
Fix: Resolve null reference exception that occurs when lang is not supplied to hanko-elements

irby:
# Description

Addresses the bug identified in #548. When lang is not supplied to hanko-elements items, the page no longer renders and a console error indicates a reference issue. According to the docs, if this attribute is not supplied, elements should fall back to the `en` language.

# Implementation

The `TranslatorProvider` is supplied `lang` and `fallbackLang` attributes. The value provided in `lang` uses `lang.toString()`. When adding in an optional chaining operator, i.e. `lang?.toString()` I saw the issue was no longer present.

# Tests

Manual testing confirmed these changes. As noted in the `Todos` section, I would like to figure out how to make an automated test to confirm these changes.

Before:

image

After (no `lang` is supplied to elements)

image

# Todos

- This was manually tested and verified by restarting the Docker containers and observing the issue was fixed. I would like to have some automated tests to verify these changes, but not quite sure how this could be achieved. I would like to make automated tests to verify:
  - When lang is supplied to hanko-elements elements with `en`, English text is displayed
  - When lang is supplied to hanko-elements elements with `de`, German text is displayed
  - When no lang is supplied to hanko-elements elements, English text is displayed


bjoern-m:
Hi @irby!

Looks good to me and would approve it as it is. I guess you could add the test cases you mentioned to the [e2e tests](https://github.com/teamhanko/hanko/tree/main/e2e). Also I am about to create unit tests for hanko-elements. It will take me some time, but they are coming (at least if I don't fail at it 🙂).

Kind regards

bjoern-m (approved):

irby:
Hi @bjoern-m if unit tests are about to be made for hanko-elements, those should be able to cover the different language scenarios. If those come around, I could make some tests based off those. Otherwise, the setup for the e2e tests would require some additional docker configurations to be passed to the quickstart (I think). I think for now, I'll leave the changes as-is. I'll try to think of some additional ways to add automated tests for this in the meantime though!

like-a-bause (approved):

PR #617, opened 2023-03-01
feat(config): allow multi word env var keys

### LINKED ISSUE #126: Backend configuration through environment variables not possible for multi-word, underline delimited keys

lfleischmann:
# Description

Fixes #126 

# Implementation

- Use [envconfig](https://github.com/kelseyhightower/envconfig) to load environment variables instead of `koanf`
    - Use[ `split_words`](https://github.com/kelseyhightower/envconfig#struct-tag-support) struct tag to correctly load vars where config keys consist of multiple words

# Tests

Added a (very) small unit test to see whether a config option is correctly overwritten by what is specified through the env var and that multivalued env vars are read correctly. To test for yourself, join config keys by `_`
(underscore), uppercase the keys, i.e. for `server.public.cors.allow_methods`
use:

```shell
export SERVER_PUBLIC_CORS_ALLOW_METHODS="GET,PUT,POST,DELETE"
```

Start the backend and observe whether your changes are correct.

### DIFF
--- a/backend/docs/Config.md
+++ b/backend/docs/Config.md
@@ -5,7 +5,7 @@ Environment variables have higher precedence than configuration via file (i.e. i
 given in the file - multivalued options, like arrays, are also _not_ merged but overwritten entirely).
 
 The schema for the configuration file is given below. To set equivalent environment variables, join keys by `_`
-(underscore) uppercase the keys, i.e. for `server.public.cors.allow_methods`
+(underscore) and uppercase the keys, i.e. for `server.public.cors.allow_methods`
 use:
 
 ```shell


like-a-bause:
fixes #126

like-a-bause (approved):

PR #631, opened 2023-03-03
chore: re-generate jsdoc

bjoern-m:
# Description  

JSDoc re-generated with jsdoc 3.6.11, where taffydb is not needed.


like-a-bause (approved):
Except for the version number, there hasn't changed anything 🤷‍♂️

bjoern-m:
Yes right, I re-generated the docs with jsdoc v3 instead of 4 where you don't need taffydb to generate the docs.

PR #653, opened 2023-03-09
feat: make conditional ui available via env var in quickstart

like-a-bause:
# Description

Adds the ability to set experimental feature conditional ui in quickstart

# Implementation

# Tests

# Todos

# Additional context

### DIFF
--- a/quickstart/main.go
+++ b/quickstart/main.go
@@ -24,6 +24,7 @@ func main() {
 		hankoUrlInternal = value
 	}
 
+	// This is handled as a "flag" if set to any value, conditional UI is enabled.
 	_, conditionalUi := os.LookupEnv("HANKO_ENABLE_CONDITIONAL_UI")
 
 	e := echo.New()
--- a/quickstart/main.go
+++ b/quickstart/main.go
@@ -24,6 +24,8 @@ func main() {
 		hankoUrlInternal = value
 	}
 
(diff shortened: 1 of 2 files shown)


bjoern-m:
Just to be sure, is it intended that the value of `HANKO_ENABLE_CONDITIONAL_UI` is discarded?  So when I set  `HANKO_ENABLE_CONDITIONAL_UI=false` it's okay that conditional UI is enabled anyway, because it's just a "flag"?

like-a-bause:
> Just to be sure, is it intended that the value of `HANKO_ENABLE_CONDITIONAL_UI` is discarded? So when I set `HANKO_ENABLE_CONDITIONAL_UI=false` it's okay that conditional UI is enabled anyway, because it's just a "flag"?

Yes. I thought of it as a boolean flag. But I don't have a strong opinion about that. We could also handle this as "must explicitly be set to true".

bjoern-m:
> > Just to be sure, is it intended that the value of `HANKO_ENABLE_CONDITIONAL_UI` is discarded? So when I set `HANKO_ENABLE_CONDITIONAL_UI=false` it's okay that conditional UI is enabled anyway, because it's just a "flag"?
> 
> Yes. I thought of it as a boolean flag. But I don't have a strong opinion about that. We could also handle this as "must explicitly be set to true".

Ok, me neither, but I think we should create a task to mention the available environment variables in the quickstart README.

bjoern-m (approved):

bjoern-m (approved):

PR #662, opened 2023-03-13
feat: user deletion

bjoern-m:
# Description

Allows users to delete their own account.

# Implementation

- Added a new API route: `DELETE /users`
- Added a new config param `config.account.allow_deletion`
- The new route is available when `config.account.allow_deletion` is set to true.
- When deletion is allowed, `<hanko-profile>` will display a "Delete account"-button.
- By clicking the button a confirmation page will be displayed

# Tests

- `frontend-sdk` tests are included.
- Tried to write a backend test, but failed once again. I assume this is because the router or middleware is not running and therefore the status code remains unchanged:

```
func (s *userSuite) TestUserHandler_Delete() {
  if testing.Short() {
    s.T().Skip("skipping test in short mode.")
  }
  err := test.LoadFixtures(s.db.DbCon, s.db.Dialect, "../test/fixtures/user")
  s.Require().NoError(err)

  tests := map[string]struct {
    userId           string
    withSessionToken bool
    expectedCode     int
  }{
    "invalid uuid": {userId: "invalid", withSessionToken: true, expectedCode: http.StatusInternalServerError},
    // ...
  }

  for name, tc := range tests {
    s.T().Run(name, func(t *testing.T) {
      e := echo.New()
      e.Validator = dto.NewCustomValidator()
      req := httptest.NewRequest(http.MethodDelete, "/user", nil)
      rec := httptest.NewRecorder()
      c := e.NewContext(req, rec)

      if tc.withSessionToken {
        token := jwt.New()
        err = token.Set(jwt.SubjectKey, tc.userId)
        s.Require().NoError(err)
        c.Set("session", token)
      }

      handler := NewUserHandler(&defaultConfig, s.storage, sessionManager{}, test.NewAuditLogger())

      _ = handler.Delete(c)
      s.Equal(tc.expectedCode, rec.Code) // => rec.Code is 200 instead of 500
    })
  }
}
```


backend/handler/user.go
```diff
@@ -220,6 +220,48 @@ func (h *UserHandler) Me(c echo.Context) error {
...
+		}
+
+		err = h.auditLogger.Create(c, models.AuditLogUserDeleted, user, nil)
+		if err != nil {
+			return fmt.Errorf("failed to write audit log: %w", err)
+		}
+
+		err = h.persister.GetUserPersisterWithConnection(tx).Delete(*user)
+		if err != nil {
+			return fmt.Errorf("failed to get user: %w", err)
```
like-a-bause, line 251:
```suggestion
			return fmt.Errorf("failed to delete user: %w", err)
```

backend/handler/user.go
```diff
@@ -220,6 +220,48 @@ func (h *UserHandler) Me(c echo.Context) error {
...
+		user, err := h.persister.GetUserPersisterWithConnection(tx).Get(userId)
+		if err != nil {
+			return fmt.Errorf("failed to get user: %w", err)
+		}
+
+		if user == nil {
+			return fmt.Errorf("unknown user")
+		}
+
+		err = h.auditLogger.Create(c, models.AuditLogUserDeleted, user, nil)
```
like-a-bause, line 244: Shouldn't we first delete the use, then write to the audit log?

bjoern-m: Yes makes sense, I changed it. But I think the `AuditLogger.Create()` method should be able to work on a transaction, because currently we cannot guarantee that the audit log will be written.

like-a-bause:
After deletion in the quickstart nothing changes in the ui. Reloading just states "Unauthorized". We don't know what the RP wants to do after account deletion. This is something to be handled by the upcoming "event system", right?

bjoern-m:
> After deletion in the quickstart nothing changes in the ui. Reloading just states "Unauthorized". We don't know what the RP wants to do after account deletion. This is something to be handled by the upcoming "event system", right?

Yes, will add profile related events in a separate PR.

like-a-bause (approved):

PR #698, opened 2023-03-25
`isready` subcommand to check if a service is ready to serve requests

### LINKED ISSUE #697: Health checks in docker images
### Checklist

### Description

I am trying to configure a [health check](https://docs.docker.com/engine/reference/builder/#healthcheck) in my Dockerfile. Here is the `test` command I use `[ "CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:9000/.well-known/jwks.json" ]` However, since Hanko base image uses a `distroless` base image, utilities like `wget`, `curl` are not present in the base image.

Output of health check:

image

How can I configure a healthcheck for my Hanko service, in this case?

Maybe we can use a base image of `alpine` instead of `distroless`. However, this might increase image size by a few MBs.

None that I know of 

### Hanko Version

v0.5.0

### Additional Context

Moreover, using `/.well-known/jwks.json` for public service's health check seems a bit hacky, it would be nice if there was a dedicated endpoint for it.

The admin service does not seem to have a health check endpoint at all.

IgnisDa:
# Description

This PR introduces the `isready` subcommand for the Hanko CLI.

Fixes #697.

# Implementation

The command takes in the service to check (`public` or `admin`) and makes an HTTP request to the corresponding health endpoint. 

# Tests

Tested manually.

image

### DIFF
--- a/backend/cmd/isready/isready.go
+++ b/backend/cmd/isready/isready.go
@@ -14,9 +14,10 @@ func NewIsReadyCommand(config *config.Config) *cobra.Command {
 		Args: cobra.OnlyValidArgs,
 		ValidArgs: []string{"admin", "public"},
 		Short: "Health check a service",
-		Long: `Checks if the specified service is healthy. Possible values are "admin" and "public".`,
+		Long: `Checks if the specified service is healthy. Possible values are "admin" and "public".
+Uses the "/health/ready" endpoint to check if the service is ready to serve requests.
+		`,
 		Run: func(cmd *cobra.Command, args []string) {
-			// check that there is only one argument
 			if len(args) != 1 {
 				fmt.Println("Please specify a service to check")
 				os.Exit(1)
--- a/backend/cmd/isready/isready.go
+++ b/backend/cmd/isready/isready.go
@@ -5,6 +5,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/teamhanko/hanko/backend/config"
 	"log"
+	"net"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


IgnisDa:
@like-a-bause This is ready to be reviewed.

backend/cmd/isready/isready.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+			service := args[0]
+			address := ""
+			if service == "admin" {
+				address = config.Server.Admin.Address
+			}
+			if service == "public" {
+				address = config.Server.Public.Address
+			}
+			if address[0] == ':' {
```
like-a-bause, line 32: You could use net.SplitHostPort() here.

backend/cmd/isready/isready.go
```diff
@@ -0,0 +1,53 @@
...
+				address = "0.0.0.0" + address
```
like-a-bause, line 33: I tried this out and it works. However to quote wikipedia:
```
RFC 1122 refers to 0.0.0.0 using the notation {0,0}. It prohibits this as a destination address in [IPv4](https://en.wikipedia.org/wiki/IPv4) and only allows it as a source address under specific circumstances.
```
We would use this as a destination here. Better to just use localhost, then.

like-a-bause (changes requested):

backend/cmd/isready/isready.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+			host, port, err := net.SplitHostPort(address)
+			if err != nil {
+				log.Fatalf("Could not parse address %s", address)
+			}
+			if host == "" {
+				host = "localhost"
+			}
+			requestUrl := fmt.Sprintf("http://%s:%s/health/ready", host, port)
+			res, err := http.Get(requestUrl)
```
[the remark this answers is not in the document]
like-a-bause: @sonatype-lift ignore

like-a-bause (approved):

PR #712, opened 2023-04-06
refactor(cmd): rework config initalization

lfleischmann:
# Description

Config initialization is done in the root command but some commands do not require a config. If these commands are executed locally in the checked out repo they work fine, but they will fail if they are executed e.g. using the Docker image. I would have to provide a config file path to the container and mount the referenced file into the container even though it is not needed for the command.

# Implementation

- Make those commands that require a config load it themselves. As a consequence:
    -  Remove `cmd.initConfig` function and `cobra.OnInitialize` call.
    -  Config validation is now part of the `config.Load` function.
- Pull down the config flag to the individual commands that require it instead of using a persistent flag on the root level (it is unintuitive to have a global flag available that, if set, on certain commands has no consequence whatsoever).

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -65,7 +65,7 @@ func Load(cfgFile *string) (*Config, error) {
 	}
 
 	if err = c.Validate(); err != nil {
-		log.Fatalf("failed to validate config: %s", err)
+		return nil, fmt.Errorf("failed to validate config: %s", err)
 	}
 
 	return c, nil


backend/config/config.go
```diff
@@ -60,6 +64,10 @@ func Load(cfgFile *string) (*Config, error) {
 		return nil, fmt.Errorf("failed to post process config: %w", err)
 	}
 
+	if err = c.Validate(); err != nil {
+		log.Fatalf("failed to validate config: %s", err)
```
FreddyDevelop, line 68: I think it is better to return the error, because the `Load` function can return an error and every checks above also return an error.

FreddyDevelop (approved):

PR #725, opened 2023-04-13
Session events

bjoern-m:
# Description

Introduces events to the `frontend-sdk`. Events are supposed to work as follows:

```typescript
import { Hanko } from "@teamhanko/hanko-frontend-sdk"

const hanko = new Hanko("https://[HANKO_API_URL]")

hanko.onSessionCreated((detail) => {
    // Executes when there already is a session, after the user signs in, or when the JWT has been updated. It will
    // work across browser windows and you can obtain the JWT from the detail object, if you need to manage it by your
    // own. Please note, that the JWT is only available, when the Hanko API configuration allows to obtain the JWT.
    console.log(`User signed in (userID: ${detail.userID}, jwt: ${detail.jwt})`);
})

hanko.onAuthFlowCompleted((detail) => {
    // Login or registration has been finished through the `<hanko-auth>` element. You can now redirect the user to a
    // secured page or fetch secured content in use of the previously issued JWT.
    console.log(`Authentication flow completed (userID: ${detail.userID})`);
})

hanko.onSessionRemoved(() => {
    // Executes across all browser windows after the session has expired. The user can now be redirected back to a
    // login page.
    console.log("User logged out or session has expired");
})

hanko.onUserDeleted(() => {
    // Executes after the user deleted the account. The user can be redirected to a "goodbye" or back to a login page.
    console.log("User deleted");
})
```

# Implementation

## Hanko-API

- API endpoints that issue a JWT also set the "X-Session-Lifetime" header field, which contains the number of seconds remaining until the JWT expires.
- The token and the WebAuthn finalisation endpoint will also return the user id, so "hanko-session-created" events can be dispatched immediately without gathering more information beforehand.

## Fronend SDK

- Functionality added to manage event listeners.
- Functionality added to manage the event dispatching.

## Elements

- Will dispatch the "hanko-auth-flow-completed" event.

# Tests

- largely included

## Additional Notes

Now that the SDK reads the JWT out of the cookie to provide it through an event detail object, it is necessary to set `session.cookie.http_only` to `false` via the Hanko-API config, when testing on localhost, because on localhost a different port counts as same-site and the cookie will be set (with http-only) by the backend. And even when `session.enable_auth_token_header` is `true` the "http-only" flag can't be changed by the frontend. When the JWT can't be read by the frontend, everything should work, only the event detail's `jwt` value is expected to be empty. But when the cookie is readable the `jwt` value should be there.


bjoern-m:
@FlxMgdnz, I created a migration guide on the fast. We should put something like this into the release notes:

# Migrating from `hanko-elements` v0.4.x to v0.5.x

- The `register()` method exported by `@teamhanko/hanko-elements` has changed:

Outdated:
```
import { register } from "@teamhanko/hanko-elements";

register({ ...options });
```

New:
```
import { register } from "@teamhanko/hanko-elements";

// The `register` function now returns an instance of the `frontend-sdk`, therefore you 
// don't need to install or import the `@teamhanko/hanko-frontend-sdk` package.
const { hanko } = await register("https://your-domain.com", { ...options });

// Get the current user
const user = await hanko.user.getCurrent();

```

- The "@teamhanko/hanko-elements" package has a new export member "createHankoClient":

```
import { createHankoClient } from "@teamhanko/hanko-elements";

// Somewhere in you code you can call `createHankoClient()` to get the `frontend-sdk` instance.
const hanko = createHankoClient("https://your-domain.com");
```

- Because you now provide the API URL with the `register()` function, you don't need to pass the value into the web components:

Outdated: 
```
<html>
  <hanko-auth api="https://your-domain.com"></hanko-auth>
  <hanko-profile api="https://your-domain.com"></hanko-profile>
</html>
```

New: 
```
<html>
  <hanko-auth></hanko-auth>
  <hanko-profile></hanko-profile>
</html>
```

- You can now add callbacks to certain events via the `frontend-sdk`:

```
import { register } from "@teamhanko/hanko-elements";

const { hanko } = await register("https://your-domain.com");

// Add a callback to be executed, when the user is logged in and ready to be redirected.
const cleanup = hanko.onSessionCreated((details) => {
	// Your code...
})

// The `cleanup()` function removes the event listener
cleanup();
```

- The "hankoAuthSuccess" event dispatched by `<hanko-auth>` will still work, but you should change your code as follow:

Outdated:
```
document.addEventListener("hankoAuthSuccess", () => {
	// Your code...
});
```

New:
```
hanko.onAuthFlowCompleted((details) => {
	// Your code...
})
```

PR #778, opened 2023-05-10
fix: set the auth cookie even when x-session-lifetime is not present

bjoern-m:
# Description

Set the auth cookie even when x-session-lifetime is not present.


FreddyDevelop (approved):
LGTM

PR #815, opened 2023-06-01
feat: introduce custom translations

bjoern-m:
# Description

- Translations can now be imported individually.
- The elements contain English translations by default.
- New `register()` options have been introduced to add translations and control the fallback language.
- See updated readme for more details.


FreddyDevelop (approved):
LGTM.

One thing I noted and I don't know if we want to include in the elements README explicitly is, that the `fallbackLanguage` also works in combination with `translationsLocation`.

bjoern-m:
> LGTM.
> 
> One thing I noted and I don't know if we want to include in the elements README explicitly is, that the `fallbackLanguage` also works in combination with `translationsLocation`.

Thanks for the feedback. I tried to improve the README, see 0c78d545c76a45d3129c42858997a19a4c364724 .

FreddyDevelop (approved):

PR #829, opened 2023-06-08
chore(hanko-elements): improve event handling for better SPA support

bjoern-m:
# Description

When multiple Hanko elements are placed on the same page, this PR prevents the elements to replay the initial event dispatching (`onSessionNotPresent` or `onSessionResumed`) after (re-)initialisation. It happened before due to a call of `hanko.relay.dispatchInitialEvents()` on the `InitPage`, after the individual component had been initialised, while reinitialisation can happen based on other events.

# Implementation

The initial state of the elements is now determined based the session detail `hanko.relay.getSessionDetail()` instead of on the `onSessionNotPresent` or `onSessionResumed` event, so there is not need to additionally call `hanko.relay.dispatchInitialEvents()` to receive one of the two events at the right time.


FreddyDevelop (approved):
LGTM

like-a-bause:
This can be closed, right? Feel free to reopen if not.

PR #862, opened 2023-06-19
refactor: use echo.HTTPError instead of redundant own type. This way …

like-a-bause:
…the prometheus mw shoul register the correct error codes.

# Description

# Implementation

# Tests

# Todos

# Additional context


FreddyDevelop (approved):
LGTM

PR #863, opened 2023-06-19
fix(examples-vue): README typo

tobihans:
# Description

Replace `React` with `Vue` inside vue example's `README.md`.


like-a-bause (approved):
Thanks for spotting this.

PR #865, opened 2023-06-20
feat(hanko-elements): add 'hidePasskeyButtonOnLogin' options

bjoern-m:
# Description

- Adds an option to the `register()` function to hide the "sign in with a  passkey" button on the login page. Whether the button is displayed or not, now depends on the `hidePasskeyButtonOnLogin` and the `enabledPasskeys` option. Enabling CUI will not hide the button, as it was the case before.
- Also refactors the "globalOptions" in the `Elements.tsx` file.

### DIFF
--- a/frontend/elements/README.md
+++ b/frontend/elements/README.md
@@ -112,15 +112,15 @@ You can also pass certain options:
 
 ```javascript
 const defaultOptions = {
-    shadow: true,                  // Set to false if you do not want the web component to be attached to the shadow DOM.
-    injectStyles: true,            // Set to false if you do not want to inject any default styles.
-    enablePasskeys: true,          // Set to false if you do not want to display passkey-related content.
+    shadow: true,                    // Set to false if you do not want the web component to be attached to the shadow DOM.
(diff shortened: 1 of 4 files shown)


FreddyDevelop:
I played around with the `hidePasskeyButtonOnLogin` and the `experimental="conditionalMediation"` setting and in some combination the UI behaves a little _strange_.

## `hidePasskeyButtonOnLogin: true` and CUI disabled
It seems strange that the divider is still shown as there are no other buttons.

Login_hpbol_true_cui_false

## `hidePasskeyButtonOnLogin: false` and CUI enabled
The divider is not shown. When a social provider (e.g. google) is enabled the divider is shown and the passkey button is shown under the divider as it should be.

Login_hpbol_false_cui_true

## `hidePasskeyButtonOnLogin: true` and CUI enabled
Don't known if this is intended behaviour (because with CUI an extra button is not needed anymore), but this behaves against the comment in the PR:

> Also the button will not disappear, when CUI has been enabled.

Login_hpbol_true_cui_true

FreddyDevelop (approved):
LGTM now

PR #883, opened 2023-07-03
fix: return shadowed error

FreddyDevelop:
# Description

Fixes that login errors are shadowed by audit log creation errors.


bjoern-m (approved):
LGTM

PR #885, opened 2023-07-03
Feat let cookie name be changed

### LINKED ISSUE #243: Allow cookie name to be changed / configured
As requested in our community, it might help with adoption when it is possible to integrate Hanko completely transparent into an existing infrastructure.

FreddyDevelop:
# Description

Allows the user to configure the cookie name.

Fixes #243 

# Implementation

The cookie name can be configured in two (three) places, the backend config with `session.cookie.name` and the `register` function of the elements (also in the frontend-sdk) depending on which component sets the cookie.

The cookie name is not exposed through the config endpoint, because then you would always have to call the config endpoint before doing a login, so that the frontend knowns the name of the cookie it should set. So I think it is better to just set the cookie name in the elements or frontend-sdk.

# Tests
## Backend cookie
- Start the quickstart app with `enable_auth_token_header: false` and `session.cookie.name: <CHOOSE_YOUR_NAME>`
- login 
- check that the cookie is set with the configured name

## Frontend cookie
- set `cookieName=<CHOOSE_YOUR_NAME>` in registerOptions in the `index.html`
- Start the quickstart app with `enable_auth_token_header: true`
- login 
- check that the cookie is set with the configured name

# Todos

Maybe add documentation on which config to use when someone wants to change the cookie name.


bjoern-m (commented):
I would rename `cookieName` to `storageKey` since it controls the name under which local storage objects are stored as well. Or you could add another option that exclusively controls the local storage key. Also the new options should be described in the README file.

frontend/frontend-sdk/src/Hanko.ts
```diff
@@ -31,60 +31,75 @@ class Hanko extends Listener {
   session: Session;
 
   // eslint-disable-next-line require-jsdoc
-  constructor(api: string, timeout = 13000) {
+  constructor(
```
bjoern-m, line 34: The docs above the class declaration needs to be updated.

frontend/frontend-sdk/src/Hanko.ts
```diff
@@ -31,60 +31,75 @@ class Hanko extends Listener {
...
     /**
      *  @public
      *  @type {Session}
      */
-    this.session = new Session();
+    this.session = new Session(options.cookieName);
   }
 }
 
+export interface Options {
```
bjoern-m, line 100: JSDoc missing

frontend/frontend-sdk/src/lib/Cookie.ts
```diff
@@ -7,7 +7,13 @@ import JSCookie from "js-cookie";
  * @subcategory Internal
  */
 export class Cookie {
-  authCookieName = "hanko";
+  authCookieName: string; // = "hanko";
```
bjoern-m, line 10:
```suggestion
  authCookieName: string;
```

frontend/frontend-sdk/src/lib/client/HttpClient.ts
```diff
@@ -129,13 +130,13 @@ class HttpClient {
   cookie: Cookie;
 
   // eslint-disable-next-line require-jsdoc
-  constructor(api: string, timeout = 13000) {
+  constructor(api: string, options: Options) {
```
bjoern-m, line 133: Same here. There are some more classes, where the `timeout` option is documented and has not been updated to the new version.

bjoern-m:
Maybe it would be beneficial to implement the use of an object for the options in the classes `Cookie`, `Session`, `Dispatcher`, `Relay` and `SessionState` as well? Instead of, for example, `new Relay(storageKey)`, we could use `new Relay({ storageKey })`, where "hanko" is the default value for `storageKey`. This way we would avoid ending up with a long list of parameters if more parameters needs to be added over time.

frontend/frontend-sdk/README.md
```diff
@@ -37,6 +37,24 @@ With a script tag via CDN:
 </script>
 ```
 
+### Options
+
+There are optional options that can be provided when creating a new `Hanko` instance.
+
+- `timeout` - The timeout (in ms) for the http requests. Default: 13000
+- `cookieName` - The cookie name under which the session token is set. Default: "hanko"
+- `storageKey` - The prefix / name of the localstorage keys. Default: "hanko"
```
bjoern-m, line 46: To make it copy & pastable I would suggest the following:
```suggestion
You can pass certain options, when creating a new `Hanko` instance:

```js
const defaultOptions = {
    timeout: 13000,      // The timeout (in ms) for the HTTP requests.
    cookieName: "hanko", // The cookie name under which the session token is set.
    storageKey: "hanko"  // The prefix / name of the localStorage keys.
};

const hanko = new Hanko("http://localhost:3000", defaultOptions);
```
```

File: frontend/elements/README.md [hunk 6]
```diff
@@ -120,6 +120,7 @@ const defaultOptions = {
                                    // from displaying any translations.
     translationsLocation: "/i18n", // The URL or path where the translation files are located.
     fallbackLanguage: "en",        // The fallback language to be used if a translation is not available.
+    cookieName: "hanko"            // The name of the cookie the session token is stored in.
```
bjoern-m, line 123: Hmm shouldn't it be the `storageKey`? Because it controls both parameters, the `cookieName` and the local storage key. Which leads me to the next question. Shall we rename `storageKey` to `localStorageKey`, where it is really the key used for storing into the local storage and whenever it means both parameters, we call it `storageKey`?

frontend/frontend-sdk/README.md
```diff
@@ -37,6 +37,24 @@ With a script tag via CDN:
...
+
+#### Example
+
+```typescript
+import { Hanko } from "@teamhanko/hanko-frontend-sdk"
+
+const hanko = new Hanko("http://localhost:3000", {
+  cookieName: "my-session-cookie"
+})
```
bjoern-m, line 55: See comment above
```suggestion
```

bjoern-m:
I think the new option interfaces should have distinct names, because in the docs, the generated HMTL file `/docs/static/jsdoc/hanko-frontend-sdk/Options.html` lists all the options called "Options". You can distiguish them by their description, but maybe this is not enough, because the individual headlines for interfaces are not equipped with an anchor. So I guess it is not possible to link to an exact option interface, only to the `Options.html`.

frontend/elements/README.md
FreddyDevelop: Yes, looks like a good idea.

frontend/frontend-sdk/src/lib/Session.ts
```diff
@@ -2,20 +2,34 @@ import { SessionDetail } from "./events/CustomEvents";
 import { SessionState } from "./state/session/SessionState";
 import { Cookie } from "./Cookie";
 
+/**
+ * Options for Session
+ *
+ * @category SDK
+ * @subcategory Session
```
bjoern-m, line 9:
```suggestion
 * @subcategory Internal
```
Otherwise it would be the only option listed under a new section "INTERFACES / SESSION".

bjoern-m (commented):
I added one last suggestion, everything else is looking good now 👍

frontend/frontend-sdk/README.md
```diff
@@ -37,6 +37,19 @@ With a script tag via CDN:
...
 
+### Options
+
+You can pass certain options, when creating a new `Hanko` instance:
+
+```js
+const defaultOptions = {
+  timeout: 13000,      // The timeout (in ms) for the HTTP requests.
+  cookieName: "hanko", // The cookie name under which the session token is set.
+  storageKey: "hanko"  // The prefix / name of the localStorage keys.
```
bjoern-m, line 48: ...and the SDK readme needs an adjustment:
```suggestion
  timeout: 13000,           // The timeout (in ms) for the HTTP requests.
  cookieName: "hanko",      // The cookie name under which the session token is set.
  localStorageKey: "hanko"  // The prefix / name of the localStorage keys.
```

bjoern-m (approved):

PR #888, opened 2023-07-04
Fix audit logger improvements

### LINKED ISSUE #683: audit: cannot make persistent log calls that adhere to atomicity of transactions
### Checklist

### Describe the bug

The audit logger currently cannot be provided with a `pop.Connection` to the underlying persister (a pattern that we use for the other `persistence` APIs) - neither at instance creation, nor when calling the `Create` method. When audit log storage is enabled and backend code requires multiple operations to be performed within the boundary of a transaction, then a call to the logger's `Create` method within that boundary does not roll back the audit log operations in case of a failure. This may lead to inconsistent audit log entries. 

### Reproducing the bug

1. Build a bogus handler, e.g.:
```go
package handler

import (
	"errors"
	"github.com/gobuffalo/pop/v6"
	"github.com/labstack/echo/v4"
	auditlog "github.com/teamhanko/hanko/backend/audit_log"
	"github.com/teamhanko/hanko/backend/persistence"
	"github.com/teamhanko/hanko/backend/persistence/models"
)

type BogusHandler struct {
	persister   persistence.Persister
	auditLogger auditlog.Logger
}

func NewBogusHandler(persister persistence.Persister, auditLogger auditlog.Logger) *BogusHandler {
	return &BogusHandler{
		persister:   persister,
		auditLogger: auditLogger,
	}
}

func (h *BogusHandler) Bogus(c echo.Context) error {
	err := h.persister.Transaction(func(tx *pop.Connection) error {
		user := models.NewUser()
		err := h.persister.GetUserPersisterWithConnection(tx).Create(user) // <- can provide connection
		if err != nil {
			return errors.New("could not create user")
		}

		err = h.auditLogger.Create(c, models.AuditLogUserCreated, &user, nil) // <- cannot provide connection
		if err != nil {
			return errors.New("could not create audit log")
		}

		return errors.New("BOOM") // <- assume the previous operations were successful, 
                                          // but at this point we introduce an error, the user creation 
                                          // is rolled back, but the audit log remains, claiming that a 
                                          // user was created, which is not correct
	})

	if err != nil {
		return err
	}

	return nil
}
```
2. Apply bogus handler in router: 
```go
func NewPublicRouter(...) *echo.Echo {
    ...
    bogusHandler := handler.NewBogusHandler(persister, auditLogger)
    e.POST("/bogus", bogusHandler.Bogus)

    return e
}
````
3. Start DB + backend
4. `curl -X POST http://localhost:8000/bogus`
5. View DB records for `users` (expected: none) and `audit_log` (expected: one `user_created` record)

### Logs

_No response_

### Configuration

```yaml
audit_log:
  storage:
    enabled: true
```

### Hanko Version

v0.5.0

### OS

None

### OS Version

_No response_

### Browser Version

_No response_

### Environment

None

### Additional Context

_No response_

FreddyDevelop:
# Description

Audit Logs can now be created within a database transaction.

Fixes #683


bjoern-m (approved):
I have tested everything I could, and the code looks good to me too. Therefore, I approve this PR.

bjoern-m (approved):

PR #929, opened 2023-07-21
fix: don't set cookie when jwt is set in header

lfleischmann:
# Description

Set session/JWT cookie only when `Session.enable_auth_token_header` is set to `false`.


FreddyDevelop (approved):
LGTM

PR #693, opened 2023-03-24
Feat user import

like-a-bause:
# Description

Adds the ability to import user data from json files.

# Implementation

# Tests

# Todos

# Additional context

PR #941, opened 2023-07-31
feat: generate more accurate backend config jsonschema

### LINKED ISSUE #930: backend: provide complete JSON schema for config
### Checklist

### Description

Having a JSON schema for the config file would improve developer experience when using a development environment that automatically checks config contents against a schema. 

The `backend/json_schema/schema_generator` already allows generating a JSON schema, but the generated schema is incomplete/not correct.

Utilize the `github.com/invopop/jsonschema` package in order to

- [use Go comments](https://github.com/invopop/jsonschema#using-go-comments) to provide schema descriptions
- [use tags](https://github.com/invopop/jsonschema#example) to, for example, properly mark properties as required/optional etc.

Automatic schema generation could be handled via CI. The schema could also be uploaded to the [JSON Schema Store ](https://www.schemastore.org/json/), for which some IDEs (JetBrains products, VSCode) have built-in support (i.e. schemas from the schema store are automatically available/selectable and retrieved from the store such that there would be no need to manually configure the generated schema in the IDE).

_No response_

### Hanko Version

any

### Additional Context

_No response_

lfleischmann:
# Description

This PR provides the possibility of generating a more accurate JSON Schema for the `backend` configuration using the `github.com/invopop/jsonschema` package which was already introduced for the purposes of building a schema for the user import through #693. 

Closes #930

# Implementation

As suggested in #930, the PR currently mainly uses [tags](https://github.com/invopop/jsonschema#example) to provide a more accurate JSON Schema. I refrained from adding more go comments, which could have been processed properly by said package, because simply using the tags already had some caveats that prevented me from going all out:

1. In order to declare properties as `optional` one has to add `omitempty` to an existing `json` tag. This is not necessarily intuitive and I currently am not sure whether this may have any impact on "internal" handling of the config file (keyword: operator). When checking for other packages providing a similar feature set (i.e. generating a schema from go code - possibly via/plus struct tags), I realized there is no package that is as well "maintained" as this one, and even then the principle of marking optional properties via `omitempty` was also used, so the decision was to keep using the invopop package.
2. Some of the more "intricate" relationships between config options cannot be easily represented using this package, especially via struct tags. For example: it is not possible to properly convey the relationship of options pertaining to the `third_party` or `rate_limiter` configs (e.g. for the former: if one of the available providers is enabled, then redirect url, error redirect url etc. must be set; for the latter: if storage type is redis, then certain redis config options must be provided) because struct properties may rely on the exact value of another struct property and this is simply not easily represented using the struct tags provided by the package. And even though it allows providing [Custom Type Definitions](https://github.com/invopop/jsonschema#custom-type-definitions), I found this way too tedious. The initial attractiveness of the package for building a schema for the backend config were the struct tags. Building your own equivalent of jsonschema-"marshal-like" functions  just seemed overkill. Long story short: I tried to represent config option relationships using struct tags wherever possible (e.g. for the `database` config, where you can/should either provide a DSN via URL or use a set dedicated properties, like host, port, dialect etc.) but wherever it was unfeasible I simply chose to rely on the existing runtime config validation.

# Tests

No new go unit tests or the like are introduced. To "test" the schema in your IDE:

- For JetBrains products, see [here](https://www.jetbrains.com/help/idea/json.html#ws_json_schema_add_custom) on how to configure JSON schemas for files.
- For VSCode, see the [yaml](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) extension and its `yaml.schemas` setting.

# Additional context
 
1. Even though #930 mentions that generating the schema could be handled by CI and that the schema could be uploaded to the [schema store](https://www.schemastore.org/json/), I am still unsure how to properly handle versioning of schemas (our config still changes rather regularly) and whether/how IDEs may handle different schema versions. For this reason I decided to not include this for now, so these tasks probably should be handled in a follow up issue/ticket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If you want to use the Hanko backend API but prefer to build your own UI, you ca
 # Contact us
 Schedule a Hanko demo. Learn how Hanko will speed up your registration and login flows with passkeys.
 
-<a target="_blank" href="https://cal.com/team/hanko/hanko-demo"><img alt="Book us with Cal.com"  src="https://cal.com/book-with-cal-light.svg" /></a>
+<a target="_blank" href="https://cal.com/team/hanko/demo"><img alt="Book us with Cal.com"  src="https://cal.com/book-with-cal-light.svg" /></a>
 
 
 # Roadmap
@@ -77,13 +77,15 @@ We are currently in **Beta** and may introduce breaking changes. Watch our relea
 |✅| OAuth logins (Sign in with Apple/Google/GitHub) |
 |✅| i18n & custom translations |
 |✅| User import |
+|⚙️| API-supported auth flows |
 |⚙️| Mobile SDKs |
+|⚙️| Basic email templating & i8n |
+| | 2FA for securing passcodes & passwords |
+| | Disable sign-ups |
+| | Refresh tokens / sessions |
 | | Sign in with Microsoft |
 
 Additional features that have been requested or that we would like to build but are currently not on the roadmap:
-- Basic email templating
-- Refresh tokens / sessions
-- 2FA for securing passcodes & passwords
 - Webhooks
 - Priviledged sessions & step-up authentication
 - Bot protection / Captcha


FreddyDevelop (approved):
LGTM. There shouldn't be any issues with the "internal" handling in our operator.

PR #943, opened 2023-08-02
fix: angular example

rishi-raj-jain:
# Description

# Implementation

# Tests

# Todos

# Additional context

### DIFF
--- a/deploy/docker-compose/todo-angular.yaml
+++ b/deploy/docker-compose/todo-angular.yaml
@@ -13,3 +13,5 @@ services:
       - "8888:8888"
     networks:
       - intranet
+networks:
+  intranet:
--- a/frontend/examples/angular/angular.json
+++ b/frontend/examples/angular/angular.json
@@ -13,6 +13,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
+            "preserveSymlinks": true,
             "outputPath": "dist/angular",
             "index": "src/index.html",
             "main": "src/main.ts",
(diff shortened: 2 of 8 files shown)


rishi-raj-jain:
While app runs fine now, the scenario 2 fails with the error as attached in the 2nd screenshot.

rishi-raj-jain:
@FlxMgdnz Did you get a look at this?

FlxMgdnz:
> @FlxMgdnz Did you get a look at this?

Thank you for your help. We'll get back to you asap.

rishi-raj-jain:
@lfleischmann 

So I figured it out! As the error said, the frontend-sdk package has the @github/webauthn-json as a devDep and misses it as a peerDep (why peer deps: https://nodejs.org/es/blog/npm/peer-dependencies#using-peer-dependencies), hence when looking into a monorepo setup where the dependency is available in the frontend folder, it's unavailable inside the individual setup.

I proceeded for now by adding:
- the @github dep as the dep to the angular project
- @github as a peer dep to the frontend-sdk package

Now the example does run fine in both the setups.

rishi-raj-jain:
Just wanna say that this for sure has been a learning.

rishi-raj-jain:
Hey there,

Did you have a look at this? Thoughts?

rishi-raj-jain:
@lfleischmann 

Thanks for pointing that out. I tried the package in a newly spin up react app without ts and then with ts. I then proceeded to check what did the tsconfig in the react example had, and the skipLibCheck: true configuration was missing from the angular's tsconfig. That worked.

The other examples have skipLibCheck: true in their tsconfig (or the ones they extend from).

I hope this all makes sense now. This looks like the cleanest fix.

rishi-raj-jain:
Okay so now, I've reversed some changes so that we can see what are the exact changes that were required. But this for sure required a deep investigation!

rishi-raj-jain:
@lfleischmann 

I'd still recommend to do it. I think it shouldn't be in this PR? but yeah I'd say do it as right now the type check is being skipped in all projects but as we ourselves know that it's a peerDep and not just a devDep for the frontend-sdk.

PR #944, opened 2023-08-02
feat: check webauthn session data expiry

lfleischmann:
# Description

Persist WebAuthn session data expiry and enforce expiry check on the server side.


FreddyDevelop (approved):
LGTM

Issue #911, opened 2023-07-17
Disable sign-ups


johnbchron:
### Checklist

- [X] I could not find a solution in the existing issues or docs.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/teamhanko/hanko/blob/main/CODE_OF_CONDUCT.md).

### Description

I would like a way to fully disable sign ups. My application involves restricting logins to only users whose emails are already inserted into the database, and it wouldn't make sense for me to allow other people to sign up. Aside from manually re-implementing the frontend components to check the db for an account before allowing sign-up, I can't find a solution that exists.

### Describe your ideal solution

A flag in the config file would suffice. Something `account.allow_creation`.

### Workarounds or alternatives

As I said above, I could re-implement or modify the existing frontend components, but that seems unnecessary.

### Hanko Version

0.8.0

FlxMgdnz:
Thanks for mentioning this! We've actually heard similar feedback in the past and are already working on planning this feature.

FlxMgdnz:
@irby Awesome! A relatively simple solution for disabling sign-ups (like @johnbchron's suggestion) is desirable, so yes, you're welcome do it. Hanko Elements would probably have to respect the setting and handle unwanted sign-up attempts gracefully. Also, the heading of <hanko-auth> ("Sign in or sign up") should probably be changed depending on the setting.

irby:
@FlxMgdnz this issue should be good to close out 😁

PR #947, opened 2023-08-03
Add configuration to disable user registration

irby:
# Description

Work to address #911, add configuration to disable user signup functionality. By default, users will be able to signup. If `account.allow_signup` is set to `false` within the configuration, it will not allow users to register from the frontend or by calling the user create API.

# Implementation

Added a new bool value to the configuration in the backend called `AllowSignup` that will be `true` by default.

On the backend, I added a check in the user create endpoint to check whether this config value is set to false. If so, it will return a 403 Forbidden response.

On the frontend, added a new page for when an account does not exist. This page does not include a button to register, it will only allow users to navigate back to the login screen. Users will only be redirected to this new screen if the config disallows user signups.

# Tests

Added a test in the backend to verify that account creation is forbidden if `account.allow_signup` is disabled in the configuration. On the frontend side, I ran the application locally to verify new site behavior of signup is disabled. I could create an e2e test to test site navigation in an automated fashion, please let me know if I have authorization to update the GitHub action.

# Todos

N/A

# Additional context

N/A

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -140,6 +140,10 @@ func DefaultConfig() *Config {
 				Interval: 1 * time.Minute,
 			},
 		},
+		Account: Account{
+			AllowDeletion: false,
+			AllowSignup:   true,
+		},
 	}
 }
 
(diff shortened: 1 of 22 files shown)


like-a-bause:
Hey Matt. Thanks a lot!
What do you mean by "let me know if I have authorization to update the GitHub action." What do you want to change? In general: Go for it, I think we have to approve such changes before anything gets executed.

irby:
@like-a-bause Added in an e2e test for the nosign flow. Ready for re-review!

FlxMgdnz:
Thanks @irby - awesome work.

irby:
Hi @FlxMgdnz , thank you very much for the tip! Greatly appreciated. How do I collect the tip?

FlxMgdnz:
Hey - you're welcome. You would have to create an account at Algora, which I can only recommend.

PR #961, opened 2023-08-10
feat: add status page

lfleischmann:


backend/handler/status.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/labstack/echo/v4"
+	"github.com/teamhanko/hanko/backend/config"
+	"github.com/teamhanko/hanko/backend/persistence"
+	"net/http"
+)
+
+type StatusHandler struct {
+	persister persistence.Persister
+	cfg       *config.Config
```
FreddyDevelop, line 12: Can be removed, it's not used anywhere.

backend/handler/status.go
```diff
@@ -0,0 +1,30 @@
...
+		persister: persister,
+		cfg:       cfg,
+	}
+}
+
+func (h *StatusHandler) Status(c echo.Context) error {
+	// random query to check DB connectivity
+	_, err := h.persister.GetJwkPersister().GetAll()
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError)
```
FreddyDevelop, line 26: I don't know how i feel about this, that on success it returns `html` and on non success it returns `json`.

It feels weird, that we return 2 different content-types depending on the status. As longer as I think about it, I think it would be better to also return `html` on the non success status, with the status shown in the html page.

FreddyDevelop (approved):
LGTM

PR #971, opened 2023-08-17
[DOCS] Add fullstack section and update nextjs guide

Ashutosh-Bhadauriya:
# Description

This PR adds a separate section for fullstack guide, nextjs guide is rewritten to make it more beginner-friendly, app directory guide and also middleware section is added.

Screenshot

### DIFF
--- a/docs/docs/guides/next.mdx
+++ b/docs/docs/guides/next.mdx
@@ -6,72 +6,113 @@ sidebar_custom_props:
   docCardIconName: nextjs-dark
 ---
 
-# Next.js
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
 
-In this guide you will learn how to use the
-[hanko-elements](https://github.com/teamhanko/hanko/blob/main/frontend/elements/README.md) web components to
-add authentication and a user profile to your Next.js application.
+# Integrate Hanko with Next.js
 
-## Install dependencies
+Learn how to quickly add authentication and user profile in your Next.js app using Hanko.
 
-Install the `@teamhanko/hanko-elements` package:
+## Install @teamhanko/hanko-elements
+
+Once you've initialized your Next app, installing hanko-elements provides you with access to the prebuilt components: `hanko-auth` and `hanko-profile`.
 
 ```shell npm2yarn
 npm install @teamhanko/hanko-elements
 ```
 
+## Add the Hanko API URL
+
+Retrieve the API URL from the [Hanko console](https://cloud.hanko.io/) and place it in your .env file.
+
+```sh title=".env"
+HANKO_API_URL=https://f4****-4802-49ad-8e0b-3d3****ab32.hanko.io
+```
+
 ## Add `<hanko-auth>` component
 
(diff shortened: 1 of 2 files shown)


docs/docs/guides/next.mdx
```diff
@@ -81,165 +122,296 @@ export default function Home() {
   )
 }
 ```
+:::info
+**Why we use dynamic import in Nextjs pages directory?** 🤔
+
+  The register function tries to set up the `<hanko-auth>` element for the browser. However, Next.js creates pages before they're shown, and it can't access the browser's tools during this process. This means our setup might not work.
+
+  To fix this, we can use Next.js's [dynamic import]((https://nextjs.org/docs/advanced-features/dynamic-import)) to load our component only when it's displayed on the client-side and skip the pre-creation step.
```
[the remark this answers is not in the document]
Ashutosh-Bhadauriya: Fixed it 👍

docs/docs/guides/next.mdx
```diff
@@ -6,72 +6,113 @@ sidebar_custom_props:
...
-element registration would fail.
+```mdx-code-block
+</TabItem>
+</Tabs>
+```
+
 
-A solution for this is to use Next's [dynamic import](https://nextjs.org/docs/advanced-features/dynamic-import) feature
-to dynamically load the component on the client-side and disable server rendering for it:
+#### Now simply import the component you just created (HankoAuth)
```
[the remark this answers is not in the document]
Ashutosh-Bhadauriya: fixed

docs/docs/guides/next.mdx
```diff
@@ -6,72 +6,113 @@ sidebar_custom_props:
...
-If you are using [Hanko Cloud](https://cloud.hanko.io), you can find the API URL on your project dashboard.
-If you are self-hosting you need to provide the URL of your running Hanko backend.
+```jsx title="components/HankoAuth.tsx" showLineNumbers
+"use client";
 
-:::
+import { useEffect } from "react";
+import { register } from "@teamhanko/hanko-elements";
+
+const hankoApi = process.env.HANKO_API_URL;
```
b4s36t4, line 47: If we're rendering this component on client side, I suppose `process.env.HANKO_API_URL` would be null. Since this is next.js app we should add prefix `NEXT_PUBLIC` here.

docs/docs/guides/next.mdx
```diff
@@ -6,72 +6,113 @@ sidebar_custom_props:
...
-    // register the component
-    // see: https://github.com/teamhanko/hanko/blob/main/frontend/elements/README.md#script
-    register(hankoApi)
-      .catch((error) => {
-        // handle error
-      });
+    register(hankoApi).catch((error) => {
+      // handle error
+    });
   }, []);
```
b4s36t4, line 76: Same but.. we don't have server side components with pages directory so this also would raise errors I feel. Can you check this as well.

Ashutosh-Bhadauriya: Hey adding NEXT_PUBLIC prefix, will work here too. thanks for pointing out

Issue #976, opened 2023-08-19
Sign in with Microsoft


FlxMgdnz:
Add Microsoft as another OAuth provider for Hanko.

sagar-kashyap:
Can you assign this to me

sagar-kashyap:
@FlxMgdnz can you assign this to me.

sagar-kashyap:
Hi @FlxMgdnz 
I am getting some difficulty while migrating. How to fix this??
`Migrator: problem creating schema migrations: couldn't start a new transaction: could not create new transaction: failed to connect to host=localhost user=postgres database=hanko: dial error (dial tcp [::1]:5432: connect: cannot assign requested address)`

b4s36t4:
@FlxMgdnz Please watch the below demo video, if you're okay with the implementation, I will raise the PR and start the documentation as well, Thanks.!
https://github.com/teamhanko/hanko/assets/59088937/653e31d2-88e2-46c1-85ee-6cffd21865b0

b4s36t4:
one key thing to note in the implementation is `email_verified` would always be true cause microsoft is not sending that key in neither of the tokens (`access_token` and `id_token`)

FlxMgdnz:
Hey @b4s36t4, this looks good 👍🏼

PR #982, opened 2023-08-20
add microsoft oauth

b4s36t4:
/claim #976

# Description
This PR add a `Microsoft` OAuth feature to the list of OAuth providers. This feature let's user to use `Microsoft` login with their backend along with the existing provides like Google, Github and Apple.

# Implementation

* Extended the available providers, created a new file `provider_microsoft.com`
* As OAuth is standard every provider does have same the config (client ID and Client Secret)
* Added Documentation inside docs folder.

## Caveats

Microsoft allows the OAuth for different sets of users, direct customers and Azure AD Customer (work and educational institutes), right now if one want to restrict the login to only one set of people it won't. Due the current config with provider we're limiting the user to go with common flow i.e which allows any user to login using microsoft.

As per the standard every OAuth should provide `email_verified` claim either in `Access Token` or `ID Token`, but microsoft doesn't provide one which caused the `EmailVerified` field set to be `true` by default.

Ref: [Link](https://learn.microsoft.com/en-us/answers/questions/812672/microsoft-openid-connect-getting-verified-email)

> I'm willing to work on the change if there is any alternative found.

# Tests

# Todos

# Additional context

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -3,6 +3,10 @@ package config
 import (
 	"errors"
 	"fmt"
+	"log"
+	"strings"
+	"time"
+
 	"github.com/fatih/structs"
 	"github.com/gobwas/glob"
 	"github.com/kelseyhightower/envconfig"
@@ -10,9 +14,6 @@ import (
 	"github.com/knadh/koanf/parsers/yaml"
 	"github.com/knadh/koanf/providers/file"
 	"golang.org/x/exp/slices"
-	"log"
-	"strings"
-	"time"
 )
 
 // Config is the central configuration type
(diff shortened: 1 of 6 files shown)


b4s36t4:
@FlxMgdnz Please review the PR!

FlxMgdnz:
Hey @b4s36t4 will you continue working on this?

FlxMgdnz:
hey @b4s36t4 we've decided to pause this until we find the time to clear some of the open questions. Thanks again for your contribution.

b4s36t4:
Hey @FlxMgdnz. So sorry for delaying the replies but I don't understand why we're closing this? Also I got rewarded for bounty which doesn't seem right because I have not completed right?

b4s36t4:
I would love to talk things before closing things without completion, can we move the conversation to discord?.

> I had health issues which made to move away from my computer for the past month.

ghstahl:
This PR also covers it.  Microsoft is an OIDC provider.
I can use it for both azureAD enterprise stuff and the social one as well.
https://github.com/teamhanko/hanko/pull/1321

ghstahl:
We may not need a microsoft social implementation on the backend.
the following generic OIDC covers it.
https://github.com/teamhanko/hanko/pull/1321

I have tested it with both Microsoft Social and AzureAD-Enterprise configurations.

Since this is generic the configuration has a setting for an image_ref, whatever that means to the front end.

PR #1016, opened 2023-09-12
Add tutorial section

Ashutosh-Bhadauriya:
# Description

Adds a new tutorials section in the sidebar


Ashutosh-Bhadauriya:
> Why is this a "tutorial" instead of a guide? Isn't Symfony also. a full stack framework?

@FlxMgdnz suggested adding this in a separate tutorials section, I think it is more of a demo app than a guide, because they seem to be doing some db stuff too, not sure though

PR #1036, opened 2023-09-22
hanko-auth component styled slightly different

Esther-Lita:
Changing the max-width  and adding padding around in the container can bring a nicer UI for the user's first impression of the Hanko components


FlxMgdnz (approved):

cayblood:
It would be nice to post a screenshot preview in PRs like this.

PR #1069, opened 2023-10-07
fix(frontend-sdk): Export types

gaurishhs:
# Description

Currently, the frontend-sdk does not export types properly, Which gives the following error: 

```
Could not find a declaration file for module '@teamhanko/hanko-frontend-sdk'. '/Users/gaurish/projects/documenzo/node_modules/@teamhanko/hanko-frontend-sdk/dist/sdk.modern.js' implicitly has an 'any' type.
  There are types at '/Users/gaurish/projects/documenzo/node_modules/@teamhanko/hanko-frontend-sdk/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@teamhanko/hanko-frontend-sdk' library may need to update its package.json or typings.ts(7016)
```

Which can be fixed by adding a dot path in types field.

# Implementation

types has to be in the exports along with the implementation for server-side

# Tests

N/A

# Todos

n/a

# Additional context
image
image

### DIFF
--- a/frontend/frontend-sdk/package.json
+++ b/frontend/frontend-sdk/package.json
@@ -8,11 +8,11 @@
   "files": [
     "dist"
   ],
-  "types": "dist/index.d.ts",
   "type": "module",
   "source": "src/index.ts",
   "exports": {
     "require": "./dist/sdk.cjs",
+    "types": "./dist/index.d.ts",
     "default": "./dist/sdk.modern.js"
   },
   "main": "./dist/sdk.cjs",


gaurishhs:
ps: the project whose url is in there has been deleted, i'm sorry.

gaurishhs:
hi yes, it worked for me

gaurishhs:
it's the same error which i get in the build logs, i dont understand the issue

FreddyDevelop:
@gaurishhs @lfleischmann I found the solution, the `default` property must be the last in the `exports` object, e.g.

```json
  "exports": {
    "require": "./dist/sdk.cjs",
    "types": "./dist/index.d.ts",
    "default": "./dist/sdk.modern.js"
  },
```

I found it [here](https://stackoverflow.com/questions/76127288/module-not-found-error-default-condition-should-be-last-one). But I can`t explain why this needs to be the last property.

gaurishhs:
maybe because default is more of a fallback thing if the above ones dont match.

PR #1082, opened 2023-10-11
added API link

krishvsoni:
# Description

I provided a direct link to the hanko API, even though the link to the official documentation is already mentioned. I added the direct API reference link so that users can conveniently access the APIs in other languages directly from the provided link. This makes it easier for users and can save them some time as well. 

image


FlxMgdnz (approved):

krishvsoni:
Happy To contribute! If the change is approved then looking forward to getting it merged ☺️

Issue #1067, opened 2023-10-06
Unable to start Email Server from test suite in parallel


irby:
### Checklist

- [X] I could not find a solution in the existing issues or docs.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/teamhanko/hanko/blob/main/CODE_OF_CONDUCT.md).

### Describe the bug

Currently, we only have a single test suite that requires an email server, passcode_test. 

For #1030 , this work will require multiple test suites to rely on an email server. The test suite setup and tear down is responsible for starting and stopping the mock email server, respectively. I've been running into problems where running multiple test suites in parallel ends up causing failures because tests cannot start / connect to the mock email server because another test suite has started / stopped it.

### Reproducing the bug

Have 2 or more test suites setup with `WithEmailServer` set to true, and with tests that handle sending an email out.

### Hanko Version

0.8.3

### OS Hanko Backend

None

### OS

None

### Environment

None

irby:
@FlxMgdnz @like-a-bause What are your thoughts on refactoring the unit tests that rely on a email server to use a Mailslurper instance? I can create an abstraction that interacts with [Mailslurper's API](https://github.com/mailslurper/mailslurper/wiki/Email-Endpoints) and use it within my tests to fetch emails / clear emails. 

I think this will come to our advantage to make it easier to inspect emails sent to the server and verify the content within.

Issue #1030, opened 2023-09-21
Email security notifications


FlxMgdnz:
Currently, Hanko backend only sends emails for passcode authentication.

Add a new config option to enable / disable account security notifications.

If enabled, send security notification emails to the user's primary email address on

- Password changed
- Email address added
- Primary email address changed (use primary email **before the change**)
- Passkey created

irby:
Would we want to configure each of these scenarios individually?

irby:
@FlxMgdnz ill take on this item

FlxMgdnz:
> @FlxMgdnz Would we want to configure each of these scenarios individually?

@irby I think it would be better to be able to toggle each warning individually, yes.

prashantug21:
@FlxMgdnz Is this issue available?

Fer-Win:
@FlxMgdnz If no is currently working on it or you would want someone else to work on it. Let me try it please..Assign it to me...

Fer-Win:
@irby ok fine, let me know if any other issue opens up

irby:
@FlxMgdnz I'm assuming emails sent out should be recorded in an audit log, correct?

Do we want to refactor the SMTP settings? Currently, these configuration settings are setup under passcode, but obviously as we include security notifications these will no longer only apply to passcodes.

FlxMgdnz:
@irby yes we should log the emails.

SMTP settings refactoring makes sense imho.

irby:
@FlxMgdnz It'll make sense to break out the SMTP settings refactor into its own PR. Do you want me to create an issue for this change, or can I link it to this issue?

FlxMgdnz:
Right, that makes sense. Linking to this issue/comment should be sufficient. Thanks @irby !

irby:
I'm willing to take myself off this ticket. I haven't actioned on this in some time. @FlxMgdnz @wrobat

Aashana101:
@FlxMgdnz kindly assign this to me

FlxMgdnz:
@irby any chance you can pick this back up?

irby:
@FlxMgdnz Sure thing. Let me sync up my fork with the latest repo and I'll take a look again at this.

irby:
Hey @FlxMgdnz wanted to get clarity on some requirements. For the following:

- Email address added
  Do we notify the email address being added and the primary email address?
- Primary email address changed (use primary email before the change)
  Are we only sending this email to the old primary email?

[...]

PR #1093, opened 2023-10-15
Fix/1067/email server parallel

irby:
# Description

This pull request resolves an issue described in PR #1067 , which will help resolve issue blocking #1030. The crux of the issue is that the current EmailServer solution is run per test suite. If multiple tests rely on an email server, it will attempt to create multiple instances of a mock SMTP server and will fail the test suite if these tests overlap. There isn't a clean way to get around this issue, even forcing tests to run non-parallel does not help, nor does ignoring errors where an existing server is already running. 

To get around this issue, I propose that we start up a separate Mailslurper Docker container that our tests will use. Mailslurper offers a good API we can use to fetch sent emails, so it helps us write assertions around emails. The downside is that this Docker instance will need to be spun up before tests are ran, so there's overhead in it. But it's the only solution I see possible for now. If alternative solutions are possible, I'd be curious to learn what they are.

In order for this test Mailslurper instance to not interfere with existing Mailslurper instances already running, the test Mailslurper container will run on different ports. i.e. 9080 / 9085 vs 8080 / 8085.

# Implementation

- Create a docker compose script in the `backend/test` directory that feeds in a configuration file defined in `backend/test/config`. This will override default web / API / SMTP ports to 9080 / 9085 / 3500, respectively. Then, I created a class that can reach out to the Mailslurper API to fetch emails sent. Using this, I've re-written the `passcode` tests to use this new class to assert on emails sent.

- Modified the go.yml PR workflow to spin up a Mailslurper Docker container before running tests, and then spin this container down after tests run.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,6 +24,15 @@ jobs:
         go generate ./...
         go build -v ./...
 
+    - name: Start test Mailslurper container
+      working-directory: .
+      run: docker-compose -f ./backend/test/docker-compose.test.yaml up --build -d
+
     - name: Test
       working-directory: ./backend
       run: go test -v ./...
+
+    - name: Stop containers
+      if: always()
+      working-directory: .
+      run: docker-compose -f ./backend/test/docker-compose.test.yaml down
--- a/backend/handler/passcode_test.go
+++ b/backend/handler/passcode_test.go
@@ -38,7 +38,7 @@ func (s *passcodeSuite) TestPasscodeHandler_Init() {
 	cfg := func() *config.Config {
 		cfg := &test.DefaultConfig
 		cfg.Passcode.Smtp.Host = "localhost"
-		cfg.Passcode.Smtp.Port = fmt.Sprintf("%d", s.EmailServer.ServicePort)
+		cfg.Passcode.Smtp.Port = fmt.Sprintf("%d", s.EmailServer.SmtpPort)
 		return cfg
 	}
 
--- a/backend/handler/passcode_test.go
+++ b/backend/handler/passcode_test.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"github.com/gofrs/uuid"
 	"github.com/stretchr/testify/suite"
 	"github.com/teamhanko/hanko/backend/config"
@@ -12,7 +13,6 @@ import (
 	"golang.org/x/crypto/bcrypt"
 	"net/http"
 	"net/http/httptest"
-	"regexp"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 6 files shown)


irby:
Hi @FlxMgdnz, yes. I could not reproduce this issue locally; I opened a PR against my own fork and [the Go test did not fail](https://github.com/irby/hanko/actions/runs/6521483320) there. I’ll continue looking to figure out why this is failing.

FlxMgdnz:
Cc @lfleischmann

irby:
Hi @lfleischmann, yes, you'll need to start a container for Mailslurper by using the command `docker-compose -f ./backend/test/docker-compose.test.yaml up --build -d` prior to running tests.

I've added documentation to the backend README on how to run these tests.

irby:
I believe I've discovered the root cause here. Because these tests are running in parallel, I believe one or more tests is (occasionally) interfering with the HTTP request to get emails from the test mailslurper instance. 

Based on the error logs in the [failed run](https://github.com/teamhanko/hanko/actions/runs/6521483409), I see that the exception was coming from Gock, a mock HTTP interceptor that's use in the [third party callback code](https://github.com/teamhanko/hanko/blob/main/backend/handler/thirdparty_callback_error_test.go).

I'm not familiar with Gock, so I don't know the best path forward, but these are my thoughts on the workaround here:

- We make the passcode_test suite non-parallel
- At the start of the TestPasscodeHandler_Init, we call `gock.Off` to ensure no HTTP interception will interfere with our Mailslurper HTTP calls.

Let me know what you think.

irby:
@lfleischmann I'll certainly take a look into it. I was curious how those test suite Docker instances we're being setup. I think it would be better to have the test suite be responsible for setting up the containers so we don't have the overhead of spinning up the docker before running tests.

irby:
@lfleischmann After using the solution you've posted, I like the solution much better. My setup is that if a suite has its `IncludeEmailServer` property is set to true, it will spin up a Mailslurper container specific to that test suite.

It's working quite well, but I'm noticing that every so often a test fails because the emails we receive from Mailslurper's API is a little out of sync with what's been sent. I'm looking into solutions to resolve this.

irby:
@lfleischmann  
> Not sure if it is the same as your "out of sync" experience but: I had to [give it some leeway](https://github.com/teamhanko/hanko/blob/main/e2e/pages/LoginPasscode.ts#L27-L29) in the e2e tests too because it looked like inbound messages were not processed "fast enough".

I was going to do something similar, as much as I don't like adding in pauses in tests, but when tests reach out to the Mailslurper API to fetch emails, I'm adding in a delay of a few milliseconds before making that call. Currently, it's a 15 ms delay. So far, I've run 30 tests without issue. I'll refactor my changes a little bit, run some more tests, and push a change up soon.

irby:
@lfleischmann This is ready for review.

PR #1097, opened 2023-10-16
Add `user export` subcommand

### LINKED ISSUE #1096: User export
In the same spirit of our existing user import feature, we need user export.

It should be in the same format / schema that we use for user import (i.e. exported Hanko data should be importable to another Hanko without any modifications):

- https://raw.githubusercontent.com/teamhanko/hanko/backend/v0.8.4/backend/json_schema/hanko.user_import.json
- https://cloud.hanko.io/examples/import_users.json

IgnisDa:
# Description

In the same spirit of our existing user import feature, we need user export.

Fixes #1096.

# Implementation

Added a method to the persister to get all `User`s. Then called it for the subcommand. 

# Tests

There's nothing much going on. I don't think we need any.

# Todos

N/A


backend/cmd/user/root.go
```diff
@@ -7,7 +7,7 @@ import (
 func NewUserCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "user",
-		Short: "User import/export(TODO) tools",
+		Short: "User import/export tools",
 		Long:  `Add the ability to import users into the hanko database.`,
```
FreddyDevelop, line 11: The long description should also reflect that there is an export function now.

IgnisDa: @FreddyDevelop done!

FreddyDevelop (approved):

PR #1129, opened 2023-10-24
docs: roadmap updates

FlxMgdnz:


FreddyDevelop (approved):
Maybe we can move the roadmap to its own page (or something different), because IMO the list gets quite long now and personally I don't like such long lists on the README.

PR #1150, opened 2023-10-30
Error message styling

Esther-Lita:
Improving the error message styling to optimise most users first look, when misconfiguring App URL on Hanko Cloud

Commit Type: chore


frontend/elements/example.css
```diff
@@ -249,7 +249,7 @@
 
 .hanko_exclamationMark {
     fill: var(--error-color, #e82020);
-    padding-right: 5px
+    padding-right: 1px 6px 0 4px
```
FlxMgdnz, line 252: Is this line correct? Shouldn't it be just "padding"?

Esther-Lita: You are right, I forgot to update to just "padding" in the example.css

frontend/elements/src/components/error/styles.sass
```diff
@@ -8,14 +8,16 @@
   color: variables.$error-color
   background: variables.$background-color
 
-  padding: .25rem
+  padding: .50rem 1.5rem .50rem .50rem
```
bjoern-m, line 11: Is it intended for the padding on the right to be three times the value of the other sides?
Screenshot 2023-11-01 at 16 08 09

frontend/elements/src/components/error/styles.sass
```diff
@@ -8,14 +8,16 @@
...
+  padding: .50rem 1.5rem .50rem .50rem 
   margin: variables.$item-margin
   min-height: variables.$item-height
 
   display: flex
-  align-items: center
+  align-items: start
   box-sizing: border-box
 
+  line-height: 20px
```
bjoern-m, line 19: Wouldn't it make more sense to set a unitless value or `normal` to avoid unexpected results?

frontend/elements/src/components/icons/styles.sass
```diff
@@ -35,7 +35,7 @@
 .exclamationMark
   @extend .icon
   fill: variables.$error-color
-  padding-right: 5px
+  padding: 1px 6px 0 4px
```
bjoern-m, line 38: Is it necessary for the icon to have left padding? This would create additional space between the padding of the error message box and the icon. Alternatively, you could adjust the padding of the error box to achieve the same effect.

frontend/elements/src/components/error/styles.sass
Esther-Lita: It is intended, That way when the error message is two lines it has a better visual, the text will break at different point.I am trying to bring it closer to the designs

frontend/elements/src/components/icons/styles.sass
Esther-Lita: You are right, and that’s definitely something I also considered. I decided to try this way in case the icon is replaced in the future or modified in size, etc the spacing could be adjusted just in the icon without influencing the rest of error message box. It is intended the icon to have 8px left so it’s not too close to the border.

frontend/elements/src/components/error/styles.sass
bjoern-m: Where the text breaks depends on many things, such as the component width, the font size and family, the language, and also there is more than one message, that can be displayed inside the box. So I don't see the point to randomly change the padding just to address one specific case.

frontend/elements/src/components/error/styles.sass
bjoern-m: And what about a unitless value that works for now? Also there actually is a shadow-part for the error text, it's called `error-text`.

frontend/elements/src/components/icons/styles.sass
bjoern-m: Ok, I can't verify this one, hence consider it as resolved. But is there a template in figma for the error box?

frontend/elements/src/components/error/styles.sass
Esther-Lita: Screenshot 2023-11-06 at 13 03 00

This error message is the first interaction with most of the new Hanko users due to an App URL misconfiguration, testing how to make it look better with the default styling we provide can be perceived as random. With your guidance I will find the right values and target the right parts to provide a better display of the error messages.

bjoern-m:
```suggestion
  padding: .5rem
```

bjoern-m (commented):
As discussed via PM, I added two suggestions.

frontend/elements/src/components/icons/styles.sass
bjoern-m: With the applied suggestions, just play around, what looks the best.

bjoern-m (approved):

PR #1251, opened 2023-12-11
fix: make samesite and domain attributes configurable for frontend sdk cookie

### LINKED ISSUE #699: frontend-sdk: cookies set via SDK not valid for subdomains
### Checklist

### Description

Cookies set by the SDK (i.e. set in response to reading the `X-Auth-Token` header in case of frontend and Hanko API running on different domains) are not valid for subdomains. If my frontend runs on `yourapp.com` and I have another API running at `api.yourapp.com` then the cookie will only be valid for `yourapp.com` but not the `api` subdomain. This is because we do not set the `domain` attribute on creating the cookie: 

```javascript
_setAuthCookie(token: string) {
    const secure = !!this.api.match("^https://");
    Cookies.set(this.authCookieName, token, { secure }); // <- no domain set
}
```

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute):

> The Domain attribute specifies which hosts can receive a cookie. If the server does not specify a Domain, the browser defaults the domain to the same [host](https://developer.mozilla.org/en-US/docs/Glossary/Host) that set the cookie, excluding subdomains. If Domain is specified, then subdomains are always included. 

I think it is a viable scenario that users run another service on such a subdomain and want to use the JWT for authentication with said service as well.

Set the domain when creating the cookie:

```javascript
_setAuthCookie(token: string) {
    const secure = !!this.api.match("^https://");
    Cookies.set(this.authCookieName, token, { secure , domain: "<DOMAIN>" }); // 
}
```

How to get the domain at this point:

1. Determine from current location (not sure how to do this in reliable/robust manner) or
2. Make it configurable and propagate via public configuration

_No response_

### Hanko Version

v0.5.0

### Additional Context

_No response_

### LINKED ISSUE #1245: Hanko cookie does not have SameSite attribute
### Checklist

### Describe the bug

When I use my Hanko-enabled site in Firefox, I get warning messages about the "hanko" cookie. This is the German message:
Das Cookie "hanko" verfügt über keinen gültigen Wert für das "SameSite"-Attribut. Bald werden Cookies ohne das "SameSite"-Attribut oder mit einem ungültigen Wert dafür als "Lax" behandelt. Dadurch wird das Cookie nicht länger an Kontexte gesendet, die zu einem Drittanbieter gehören. Falls Ihre Anwendung das Cookie in diesen Kontexten benötigt, fügen Sie bitte das Attribut "SameSite=None" zu ihm hinzu. Weitere Informationen zum "SameSite"-Attribut finden Sie unter https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite.

### Reproducing the bug

Use Hanko auth in Firefox, check the Console

### Logs

_No response_

### Configuration

_No response_

### Hanko Version

0.9.1

### OS Hanko Backend

None

### OS Version Hanko Backend

_No response_

### OS

None

### OS Version

_No response_

### Browser Version

_No response_

### Environment

None

### Additional Context

_No response_

lfleischmann:
# Description

- The SameSite attribute is not set for cookies set by the SDK. 
- The cookie will only be sent for requests to the exact same domain but not to subdomains.

Fixes #1245 
Fixes #699 

# Implementation

This PR makes both the SameSite and the Domain attribute configurable in the `frontend-sdk` via options when instantiating the main client instance and via the `elements` options when `register`-ing components.

# How to test

1. Modify `/etc/hosts` file, e.g. add 

```
127.0.0.1 myapp.local
127.0.0.1 api.myapp.local
```

2. Modify `frontend/examples/express` files: :
	- In `server.js` change `corsOptions.origin` from `localhost` to `myapp.local` : 
		```
		const corsOptions = {
			origin: "http://myapp.local:8888",
			...,
		};

		```
	- In `.env`, change `HANKO_API_URL` to the API URL of a Hanko Cloud instance.
3. Select one of the example applications, e.g. React. In that case, modify `frontend/examples/react` files:
	- In `.env`, change `REACT_APP_HANKO_API` to the API URL of a Hanko Cloud instance and `REACT_APP_TODO_API` to `http://api.myapp.local:8002`.
4. Set the correct API URL in your Hanko Cloud project, i.e. `http://myapp.local:8888`.
5. Run `npm run build` in the `frontend` directory.
6. Run `npm run start` for both the `frontend/examples/express` and the `frontend/examples/react` app.
7. Modify `options` passed to `register` calls in `frontend/examples/react` (i.e. in `HankoAuth`/`HankoProfile`).

# Todos/Issues/Questions

- Should the SameSite attribute even be configurable, or should we just set it to a default?

### DIFF
--- a/frontend/elements/src/Elements.tsx
+++ b/frontend/elements/src/Elements.tsx
@@ -106,9 +106,7 @@ export const register = async (
     fallbackLanguage: "en",
     ...options,
   };
-  console.log(
-    `ELEMENTS REGISTEROPTIONS COOKIE DOMAIN: ${options.cookieDomain}`,
-  );
+
   globalOptions.hanko = new Hanko(api, {
     cookieName: options.storageKey,
     cookieDomain: options.cookieDomain,


frontend/elements/src/Elements.tsx
```diff
@@ -104,9 +106,13 @@ export const register = async (
     fallbackLanguage: "en",
     ...options,
   };
-
+  console.log(
+    `ELEMENTS REGISTEROPTIONS COOKIE DOMAIN: ${options.cookieDomain}`,
+  );
```
FreddyDevelop, line 111:
```suggestion
```

frontend/frontend-sdk/src/lib/client/HttpClient.ts
```diff
@@ -114,11 +114,13 @@ class Response {
  * @subcategory Internal
  * @property {number} timeout - The http request timeout in milliseconds.
  * @property {string} cookieName - The name of the session cookie set from the SDK.
+ * @property {string=} cookieDomain - The domain where cookie set from the SDK is available. Defaults to the domain of the page where the cookie was created.
  * @property {string} localStorageKey - The prefix / name of the local storage keys.
  */
 export interface HttpClientOptions {
   timeout: number;
   cookieName: string;
+  cookieDomain?: string;
```
FreddyDevelop, line 123: Is this one needed? I did not find any usage of it. (Or am I blind 🙈?)

FreddyDevelop (approved):

PR #1255, opened 2023-12-13
fix(flowpilot): email regex match on empty or nil input value

lfleischmann:
# Description

Input validation currently attempts a regex match on email input values even if the input is optional but empty or nil. The match fails and validation returns an error.

# Implementation

These changes fix it by first checking if the input is optional and whether it has an empty or nil value.

### DIFF
--- a/backend/flowpilot/input.go
+++ b/backend/flowpilot/input.go
@@ -242,7 +242,7 @@ func (i *DefaultInput) validate(stateName StateName, inputData ReadOnlyActionInp
 		}
 	}
 
-	if i.dataType == EmailType && !isRequired && !hasEmptyOrNilValue {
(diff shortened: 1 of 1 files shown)


backend/flowpilot/input.go
```diff
@@ -239,7 +242,7 @@ func (i *DefaultInput) validate(stateName StateName, inputData ReadOnlyActionInp
 		}
 	}
 
-	if i.dataType == EmailType {
+	if i.dataType == EmailType && !isRequired && !hasEmptyOrNilValue {
```
bjoern-m, line 245: The value should be matched against the regex when `isRequired` is `true` as well.

bjoern-m (approved):

PR #1263, opened 2023-12-15
fix(flow-api): register login identifier 

lfleischmann:


backend/flow_api/flow/registration/action_register_login_identifier.go
```diff
@@ -118,15 +118,13 @@ func (a RegisterLoginIdentifier) Execute(c flowpilot.ExecutionContext) error {
 		if deps.Cfg.Password.Enabled {
 			return c.StartSubFlow(passcode.StatePasscodeConfirmation, StatePasswordCreation)
 		} else if !deps.Cfg.Passcode.Enabled || (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool()) {
-			return c.StartSubFlow(passcode.StatePasscodeConfirmation, passkey_onboarding.StateOnboardingCreatePasskey, shared.StateSuccess)
+			return c.StartSubFlow(passkey_onboarding.StateOnboardingCreatePasskey, shared.StateSuccess)
 		}
 	} else if deps.Cfg.Password.Enabled {
 		return c.ContinueFlow(StatePasswordCreation)
-	} else if !deps.Cfg.Passcode.Enabled {
+	} else if !deps.Cfg.Passcode.Enabled || (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool()) {
```
FreddyDevelop, line 125: Doesn't it have to be like this: 
```suggestion
	} else if !deps.Cfg.Passcode.Enabled && (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool()) {
``` 
otherwise the user can still end up in `StateOnboardingCreatePasskey` even when the webauthn API is not available or onboarding is disabled as only passcodes have to be disbaled.

backend/flow_api/flow/registration/action_register_login_identifier.go
```diff
@@ -118,15 +118,13 @@ func (a RegisterLoginIdentifier) Execute(c flowpilot.ExecutionContext) error {
 		if deps.Cfg.Password.Enabled {
 			return c.StartSubFlow(passcode.StatePasscodeConfirmation, StatePasswordCreation)
 		} else if !deps.Cfg.Passcode.Enabled || (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool()) {
-			return c.StartSubFlow(passcode.StatePasscodeConfirmation, passkey_onboarding.StateOnboardingCreatePasskey, shared.StateSuccess)
+			return c.StartSubFlow(passkey_onboarding.StateOnboardingCreatePasskey, shared.StateSuccess)
```
FreddyDevelop, line 121: Doesn't it have to be like this: `!deps.Cfg.Passcode.Enabled && (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool())` otherwise the user can still end up in `StateOnboardingCreatePasskey` even when the webauthn API is not available or onboarding is disabled as only passcodes have to be disabled.

FreddyDevelop (commented):
When I am not mistaken we have another case which would be not handled (when my comments are applied):

- identifier: `email`, `username` (`username` required)
- `passcode.enabled: true`
- `password.enabled: false`
- webauthn is available on the user device

A user now only enters a username. Then the user should be forwarded to `StateOnboardingCreatePasskey` so that the user can create at least one credential. But right now he would get `ErrorFlowDiscontinuity` because there is no path handling that case (after my comments are applied).

backend/flow_api/flow/registration/action_register_login_identifier.go
```diff
@@ -110,23 +110,21 @@ func (a RegisterLoginIdentifier) Execute(c flowpilot.ExecutionContext) error {
...
 		if err := c.Stash().Set("passcode_template", "email_verification"); err != nil {
 			return fmt.Errorf("failed to set passcode_template to stash: %w", err)
 		}
 
 		if deps.Cfg.Password.Enabled {
 			return c.StartSubFlow(passcode.StatePasscodeConfirmation, StatePasswordCreation)
-		} else if !deps.Cfg.Passcode.Enabled || (deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool()) {
+		} else if deps.Cfg.Passkey.Onboarding.Enabled && c.Stash().Get("webauthn_available").Bool() {
 			return c.StartSubFlow(passcode.StatePasscodeConfirmation, passkey_onboarding.StateOnboardingCreatePasskey, shared.StateSuccess)
 		}
```
FreddyDevelop, line 122: When passwords and passkey onboarding are disabled, an `ErrorFlowDiscontinuity` would be returned instead of just going strait to passcode confirmation and after that to success. So I think an `else` block with `c.StartSubFlow(passcode.StatePasscodeConfirmation, shared.StateSuccess)` is missing.

FreddyDevelop: It will be an "email verification" as the else block should be inside the `if email != "" && deps.Cfg.Identifier.Email.Verification` block.

FreddyDevelop (approved):

PR #1270, opened 2023-12-20
Feat flowpilot links

FreddyDevelop:
# Description

Adds the ability to return links in the flowpilot response. This way we do not need to change our existing third party implementation but can use static links for it. We also gain the possibility to add for example legal links (e.g. link to privacy policy) and can show them in elements.

# Implementation

Hooks have now a function in their context to add links. Each link consist of a `name`, `href`, `category` and `target`. 
For now, a link can be in one of 3 categories: `oauth`, `legal` and `other`. Links for login through a third party should be in the `oauth` category. Privacy policy, terms of service links should be in the category `legal`, any other links in `other`.

The `target` is the html `target` property and can be on of its defined values (e.g. `_blank`)

Currently only links for third party login are implemented. Other links can be added quite easily.

# Tests

Configure at least one third party provider and check if you can login/register by following the link.

### DIFF
--- a/backend/flow_api/flow/shared/hook_generate_oauth_links.go
+++ b/backend/flow_api/flow/shared/hook_generate_oauth_links.go
@@ -46,8 +46,9 @@ func (h GenerateOAuthLinks) generateHref(c echo.Context, provider string, return
 	host := c.Request().Host
 	forwardedProto := c.Request().Header.Get("X-Forwarded-Proto")
 	if forwardedProto == "" {
-		// If X-Forwarded-Proto is not set, set it to 'http' and hope that an HTTPS redirect is in place.
-		// It cannot be set to 'https' because local testing would be more complicated as you would need to set up https.
+		// Assume that a proxy is setting the X-Forwarded-Proto header correctly. Hanko should always be deployed behind a proxy,
+		// because you cannot start the backend with https and passkeys only work in a secure context.
+		// If the X-Forwarded-Proto header is not set, set it to 'http' because otherwise you would need to set up a https environment for local testing.
 		forwardedProto = "http"
 	}
 


backend/flowpilot/link.go
```diff
@@ -0,0 +1,84 @@
...
+// Link defines the interface for links.
+type Link interface {
+	Target(LinkTarget) Link
+	Href(string) Link
+
+	toPublicLink() PublicLink
+}
+
+// DefaultLink represents a link with its options.
+type DefaultLink struct {
```
bjoern-m, line 33: I think the struct type `DefaultLink` can be private

backend/flow_api/flow/shared/hook_oauth_links.go
```diff
@@ -0,0 +1,63 @@
+package shared
+
+import (
+	"fmt"
+	"github.com/labstack/echo/v4"
+	"github.com/teamhanko/hanko/backend/flowpilot"
+	"net/url"
+)
+
+type OAuthLinks struct {
```
bjoern-m, line 10: I would like to suggest adding a verb to the hook action name, e.g `GenerateOAuthLinks` (instead of `OAuthLinks`)

backend/flow_api/flow/shared/hook_oauth_links.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	return nil
+}
+
+func (h OAuthLinks) generateHref(c echo.Context, provider string, returnToUrl string) string {
+	host := c.Request().Host
+	forwardedProto := c.Request().Header.Get("X-Forwarded-Proto")
+	if forwardedProto == "" {
+		// if X-Forwarded-Proto is not set, set it to http and hope that a https redirect in place
```
bjoern-m, line 49:
```suggestion
		// If X-Forwarded-Proto is not set, set it to 'http' and hope that an HTTPS redirect is in place.
```

...but the comment within the helper function is difficult to understand without knowledge of the context in which it is used. And does it mean, that the function only works, when a redirect happened?

FreddyDevelop: Well, this function is to generate the Href of a link. And to minimise magic in the frontend part (e.g. elements) this function should also add the domain to the links, but in order to do so it needs the domain and the scheme. The domain is easy to get on, as we can get it from the `echo.Context` (`c.Request().Host`). But to know the scheme it is more difficult. Because we do not provide a way of starting Hanko with TLS, Hanko needs to be deployed behind a proxy (e.g. nginx) which does TLS termination. The proxy can send the `X-Forwarded-Proto` header which contains the scheme we need, but it can be empty. And when it is empty we have no way to know which scheme to use. 

So to answer your question, this helper function will work correctly when the proxy sends the `X-Forwarded-Proto` header or when your proxy forces an https redirect.

I can try to rewrite the comment, so that it more clear.

bjoern-m: I believe it would be clearer to state that you assume the proxy has set the header, rather than relying on the hope that a redirect is in place.

bjoern-m (approved):

Issue #692, opened 2023-03-22
Webhooks


FlxMgdnz:
Hanko's elements currently feature an event system to inform the surrounding app about certain events, e.g. when the user was successfully authenticated.

But there may be situations where it would be desired to also allow services that are not directly connected to the frontend element to react to certain events, e.g. a changed email address or deleted account.

For that, we could draft and implement a basic webhook system.

ghstahl:
I would look at this opensource project.
https://www.svix.com/
Its got all the features my other company needed.
Not using it though because of the CONS below.  Too bad, its nice.  


## Pros: 

Stable, good feature set

## Cons: 

Written in Rust (really)!  This limits the pool of contributors too much.  
Postgres, really would have liked this project to have the repository pattern so that other databases can be used without needed to change the engine.
Writes everything(success and error), which is great for tracking, not so great when the load going through is massive.  
There needs to be some means to configure some webhooks to be silent on success and issue errors to end users that a message couldn't get through.
A feature to send a tracer through the system would be nice.

Issue #1051, opened 2023-09-30
Admin API: Update user


FlxMgdnz:
Expose functionality via the admin API to update user data.

- Add email
- Change primary email
- Delete email

Let's not touch passkeys and passwords for now.

cayblood:
One thing we would want is for this API to automatically send a confirmation code to any new email addresses, with webhook notifications for email status (unconfirmed, confirmed).

IAmRiteshKoushik:
Hey, I would like to take up this issue. Could I get some directions on how to go about it ? Relevant files and folders perhaps ?

FlxMgdnz:
> One thing we would want is an option for this API to automatically send a confirmation code to any new email addresses, with webhook notifications for email status (unconfirmed, confirmed).

To refine this a bit more I'd like to understand what's the use case you want to support @cayblood ?

cayblood:
The main concern is that if we can attach new emails to users programmatically, we need a way of making sure that the email we're attaching is valid. Otherwise we might change a user's email but set it to an email that cannot really receive mail. So having the API automatically verify any email addresses we want to add saves us the trouble of building a separate mechanism for doing something that Hanko has already implemented for other workflows.

FlxMgdnz:
Got it, thanks @cayblood. So when adding an email address via admin API, there should be an option to trigger a verification email.

We'd need to implement a verification link for this non-user-initiated scenario though, as a passcode (what we currently do for email verification) won't work here.

We don't have webhooks yet either, but a webhook notification for when the verification happens seems reasonable. This would make sense as well for configurations where email verification is not required on account creation.

cayblood:
I suppose it would be possible to implement this without built-in verification with the warning that developers are responsible for verifying any emails they add.

FlxMgdnz:
That would keep things simple for now 👍🏼 

We're tracking verification links in #1063. As soon as we have them we can think about adding them as an option to certain admin actions.

proitconsult:
I'm also interested in this feature. I think it is a sensible trade-off to make sure the new email address has been validated in some way before using the API to update it on behalf of a user.

shaneHowearth:
Hi

I have a couple of questions regarding the possible behaviour of this change.

If an email is deleted, and it's the primary email for a user, what is the expected behaviour (I'm expecting an error preventing the user from being left in a state where it does not have a primary account maybe?)

[...]

PR #1291, opened 2024-01-12
feat(webhooks): add webhooks

shentschel:
# Description

This PR implements webhooks into Hanko. The primary idea is to provide users with the ability to configure webhooks for their applications. For this the user can either use the config file or a database webhook. You can enable or disbale single database webhooks. They are also disabled after reaching a critical error count (currently > 5) or their expiration date (30 days from last usage)

Closes: #692

# Implementation

webhooks can be configured either via config file (do not expire) or via the new admin api endpoints for webhooks (database webhooks with expiration and failure counter)

Webhooks are chained together so that someone can either subscribe to a main event (e.g. User) and gets all updates or to a sub event (e.g. user.create) and gets only all events which are chained to the sub event.

The current implementation uses one buffered go channel with a buffer size = len(registeredWebhooks) and one worker thread. Here we should also do an extensive performance check father in the future.

the following events are currently implemented:

* user
  * user.create
  * user.delete
  * user.update
    * user.update.email
      * user.update.email.create
      * user.update.email.primary
      * user.update.email.delete  

A webhook also sends its data encrypted as JWT. The JWT key used is the same as the one for session JWTs.

# Tests

For a test you need a reachable url. For my test I extended quickstart with a listener to /hook and used the docker compose url for quickstart to configure a webhook for multiple events. The endpoint did nothing except logging the incoming request

# Todos

* [x] Update Spec and JSON schema for config/api
* [x] Add Tests
* [x] Add Docs
* [x] Add #1051


backend/webhooks/worker.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (w *Worker) triggerWebhook(job Job) error {
+	now := time.Now()
+	// check for expire date
+	err := job.Hook.DisableOnExpiryDate(now)
+	if err != nil {
+		return err
+	}
+
+	if !job.Hook.IsEnabled() {
```
FreddyDevelop, line 50:
```suggestion
	if job.Hook.IsEnabled() {
```

backend/webhooks/config_hook.go
```diff
@@ -0,0 +1,37 @@
...
+func (ch *ConfigHook) DisableOnFailure() error {
+	return nil
+}
+
+func (ch *ConfigHook) Reset() error {
+	return nil
+}
+
+func (ch *ConfigHook) IsEnabled() bool {
+	return false
```
FreddyDevelop, line 36:
```suggestion
	return true
```

backend/webhooks/manager.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+}
+
+func (m *Manager) GenerateJWT(data interface{}) (string, error) {
+	issuedAt := time.Now()
+	expiration := issuedAt.Add(5 * time.Minute)
+
+	token := jwt.New()
+	_ = token.Set(jwt.SubjectKey, data)
```
FreddyDevelop, line 117: The JWTs never have the `data` as subject key in them, so there must be something wrong here.

FreddyDevelop (commented):
There is some stuff that needs to be changed so that the webhooks will work properly. 

I don't like what data is added to the JWT, because when I subscribe to the `user` event, then I get different data for each event, e.g. when a user is created, I get the newly created user, when an email is created I get the new emailID, but there is no way of knowing, which data I get, as the event id is `user`. 

IMO the `user`, `user.create`, `user.update`, `user.delete` should contain the complete user object and `user.update.email.create`, `user.update.email.primary`, `user.update.email.delete` should contain the complete email object with a reference to the user.

But I don't know what to do with `user.update.email`, because you don't know if an email was created or deleted as you only get `user.update.email` as indicator. Maybe we should add the specific event type to the JWT instead of the one which is registered. So when someone has registered the `user` event, and an email is created, then the webhook body contains `user.update.email.created` instead of `user`.

Also a webhook should only get triggered once per action, currently when someone registers the `user.update.email` and `user.update.email.create` event on one webhook, the webhook will get called twice when an email is created.

backend/webhooks/manager.go
shentschel: I learned something new: jwt.Subject can only be string. So I'll change it to `"hanko"` and add a data key which contains the event data

shentschel:
Now with tests and update for all review findings :)

backend/persistence/models/webhook.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/gofrs/uuid"
+)
+
+// Webhook is used by pop to map your webhooks database table to your go code.
+type Webhook struct {
+	ID            uuid.UUID     `json:"id" db:"id"`
+	Callback      string        `json:"callback" db:"callback"`
+	Enabled       bool          `json:"enabled" db:"enabled"`
+	Failures      int           `json:"failures" db:"failures"`
+	ExpiresAt     time.Time     `json:"expires_at" db:"expires_at"`
```
FreddyDevelop, line 18: After reviewing the PR a second time, I was wondering why should a Webhook expire after some time? 
Because it is always frustrating, when you have a project which is not used much and after some time something does not work correct because there is something expired because of little usage.

backend/webhooks/database_hook_test.go
FreddyDevelop, line 1: The `test.NewPersister` method should not be used anymore,it is complicated to use, because every persister must be included. But we forgot to mark it as deprecated.

Please use `test.Suite` with fixtures, which uses a real database in the background.

backend/webhooks/manager_test.go
FreddyDevelop, line 1: Here also please use `test.Suite` instead of `test.NewPersister`.

backend/webhooks/events/events.go
```diff
@@ -0,0 +1,44 @@
+package events
+
+import "github.com/teamhanko/hanko/backend/persistence/models"
+
+type Event string
+
+const (
+	User         Event = "user"
```
FreddyDevelop, line 8: This event is only used in tests and has no use anywhere else so it should be removed.

backend/webhooks/events/events.go
```diff
@@ -0,0 +1,44 @@
...
+	UserCreate   Event = "user.create"
+	UserUpdate   Event = "user.update"
+	UserDelete   Event = "user.delete"
+	Email        Event = "user.update.email"
```
FreddyDevelop, line 12: This event is only used once (while checking if an event is valid), so it should be removed.

FreddyDevelop:
While testing, the webhooks do trigger falsely, because when email verification is enabled, the webhook (user.update.email.create) is triggered before the email verification is completed. So the webhooks are not that useful, because when you want to update your local database you do not get the correct user info (new email is missing). Also the webhook would be called even when the verification was canceled or not successful.

backend/persistence/models/webhook.go
shentschel: That was a security best practice I found. It prevents data leakage in case someone else buys the domain after an project was abandoned. I'm not hard set on keeping the expire mechanism. Alternatively we can introduce a config parameter to enable/disable the expiring mechanism.

shentschel:
I (hopefully) fixed all findings except the expire mechanism. There we need to decide if we:

* want to keep it the way it is
* remove it
* add an option to enable / disable it

FreddyDevelop:
I think should add an option to enable / disable the expire mechanism.

backend/test/persister.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/teamhanko/hanko/backend/persistence/models"
 )
 
+// Deprecated: NewPersister is deprecated. User test.Suite instead
```
FreddyDevelop, line 9:
```suggestion
// Deprecated: NewPersister is deprecated. Use test.Suite instead
```

backend/webhooks/events/events.go
```diff
@@ -23,7 +21,7 @@ func StringIsValidEvent(value string) bool {
 func IsValidEvent(evt Event) bool {
 	var isValid bool
 	switch evt {
-	case User, UserCreate, UserUpdate, UserDelete, Email, EmailCreate, EmailPrimary, EmailDelete:
+	case "user", "user.update.email", UserCreate, UserUpdate, UserDelete, EmailCreate, EmailPrimary, EmailDelete:
```
FreddyDevelop, line 24: Why are `user` and `user.update.email` valid webhook events? When I create a webhook with e.g. `user` this never gets called, or am I missing something?

shentschel: it will. You suggested that if we trigger a webhook e.g. user.create that it should also be triggered for all people who subscribed to user (nested events). Did I misunderstand you on that one?

FreddyDevelop: Ohh sorry, I think I confused myself. You understood everything correct. But now as I am not confused anymore, I think I made a mistake by asking to remove the declaration of theses values https://github.com/teamhanko/hanko/pull/1291#discussion_r1465187247. They should be added back, so we have an easy to access list which events are supported.

shentschel: They are back and the option for disabling time based expiration is also implemented

backend/handler/email.go
```diff
@@ -138,6 +140,10 @@ func (h *EmailHandler) Create(c echo.Context) error {
 			return fmt.Errorf("failed to create audit log: %w", err)
 		}
 
+		if !h.cfg.Emails.RequireVerification {
+			utils.NotifyUserChange(c, tx, h.persister, events.EmailCreate, userId)
+		}
+
```
FreddyDevelop, line 146: We also need to check if the user has at least one email, because when a new user registers himself, we get `user.update.email.create` (after email verification is done) which is the wrong event, we want `user.create`. So if a user already has an email trigger `user.update.email.create`, if the user has no email trigger `user.create`.

backend/handler/passcode.go
```diff
@@ -377,6 +382,19 @@ func (h *PasscodeHandler) Finish(c echo.Context) error {
 			return fmt.Errorf("failed to create audit log: %w", err)
 		}
 
+		// notify about email verification result. Last step to prevent a trigger and rollback scenario
+		if h.cfg.Emails.RequireVerification && wasUnverified {
+			var evt events.Event
+
+			if len(user.Emails) >= 1 {
```
FreddyDevelop, line 389: Unfortunately this check does not work, because `user.Emails` is set before, when the user has no emails https://github.com/teamhanko/hanko/blob/38209aa1bdf4c40edf634455594bb899e821e86c/backend/handler/passcode.go#L348 and because of that the check is always `true`.

shentschel: should be fixed now. I moved the check to the same level as `wasUnverified` and am only using the result for the trigger if else

FreddyDevelop:
Since you added new config options, can you run `backend/json_schema/schema_generator.go`, so that the json schema gets updated? Otherwise everything else looks good now.

shentschel:
Seems that I had done it before. A run of `backend/json_schema/schema_generator.go` produced no change. But I also used the time to update the branch with all main changes :)

FreddyDevelop (approved):

PR #1316, opened 2024-02-01
feat: extend third-party account linking

lfleischmann:
# Description

Introduce a per provider configuration option that allows users to define whether automatic linking of accounts on sign-up/sign-in is enabled or disabled.

# Implementation

When `allow_linking` is set to `true` for a provider the account linking logic now links (i.e. creates a new `identity`) accounts if 

- there is no `identity` for the given provider user ID (otherwise it counts as a sign-in) and 
- there is an existing user with the email address used at the provider (otherwise it counts as a sign-up)

# How to test

1. Configure client with at least 2 providers of your choice.
2. Set `allow_linking` for each of the providers to your liking and start necessary applications (`quickstart` is probably easiest, remember to change the config in the `deploy/docker-compose` directory if you do so)
3. The two new main scenarios to test out are:
    - Create a user/an account with Hanko (don't use OAuth for sign-up), log out and _then_ do an OAuth login with one of your configured providers. The primary e-mail address at the provider should be the same as the address of the Hanko user you previously created. There should now be a link/connection to the provider via said e-mail address. 
    -  Use the user from the scenario above and sign in with another one of your configured providers. The same e-mail address should now be linked to both/two providers.

### DIFF
--- a/backend/json_schema/hanko.config.json
+++ b/backend/json_schema/hanko.config.json
@@ -678,8 +678,7 @@
           "type": "string"
         },
         "allow_linking": {
-          "type": "boolean",
-          "default": true
+          "type": "boolean"
         }
       },
       "additionalProperties": false,
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -9,9 +9,9 @@ import (
 	"github.com/knadh/koanf"
 	"github.com/knadh/koanf/parsers/yaml"
 	"github.com/knadh/koanf/providers/file"
+	zeroLogger "github.com/rs/zerolog/log"
 	"github.com/teamhanko/hanko/backend/ee/saml/config"
 	"golang.org/x/exp/slices"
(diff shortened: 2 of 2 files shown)


backend/json_schema/hanko.config.json
```diff
@@ -675,14 +675,19 @@
...
         }
       },
       "additionalProperties": false,
       "type": "object",
       "required": [
         "enabled",
         "client_id",
-        "secret"
+        "secret",
+        "allow_linking"
```
FreddyDevelop, line 690: I'am confused, the default value of `allow_linking` (currently) is set to `true` for all providers, why is this a required field?

FreddyDevelop (approved):

FreddyDevelop (approved):

FreddyDevelop (approved):

PR #1357, opened 2024-02-20
fix(flow-api): profile after concurrency fix

lfleischmann:
Allow hooks to be run before each action.
Introduce Finalize interface method for actions.


bjoern-m (approved):
LGTM

PR #1371, opened 2024-02-29
feat(flow-api): add audit logs

lfleischmann:
# Description

Adds audit logs. Also fixes the issue of passkeys not being persisted during "onboarding" on login.

# Implementation

- `audit_logs` DB table has a new column `details` for recording additional information.
- Details can now be passed to the methods of the audit logger as functional options.
- Audit logger methods have been refactored to allow masking/obfuscating log details for both the database logs and the console logs. Masking can be toggled via config option `audit_log.mask`. 

# Additional context

- Masking/obfuscating is currently done via external library, not sure if how it does this is really the way we want it to work so we may need to exchange it with a custom implementation.
- As we agreed, `thirdparty` audit logs have not been changed yet.

### DIFF
--- a/backend/flow_api/flow/shared/hook_email_persist_verified_status.go
+++ b/backend/flow_api/flow/shared/hook_email_persist_verified_status.go
@@ -17,7 +17,7 @@ func (h EmailPersistVerifiedStatus) Execute(c flowpilot.HookExecutionContext) er
 	deps := h.GetDeps(c)
 
 	if !c.Stash().Get("email_verified").Bool() {
-		return errors.New("no email has been confirmed as verified")
+		return nil
 	}
 
 	if !c.Stash().Get("email").Exists() {


FreddyDevelop (commented):
I don't like the masking library, it states that only the first three letters of the email and the domain are kept, but when I input a longer email address like `this_is_a_very_long_email_to_test_masking_2@hanko.io` it only masks a very small potion of it: `thi****_a_very_long_email_to_test_masking_2@hanko.io` It should not only mask 4 characters, It should mask every character starting with the fourth except for the domain part.

When emailVerification is disabled and I want to verify my email in the profile, I get a passcode, but after the passcode was verified, the email is still unverified. When I then enabled emailVerification in the config and verify the email, it works. When I manually want to verify my email, it should work.

backend/flow_api/flow/shared/hook_email_persist_verified_status.go
```diff
@@ -14,62 +16,93 @@ type EmailPersistVerifiedStatus struct {
 func (h EmailPersistVerifiedStatus) Execute(c flowpilot.HookExecutionContext) error {
 	deps := h.GetDeps(c)
 
-	if !deps.Cfg.Emails.RequireVerification {
-		return nil
+	if !c.Stash().Get("email_verified").Bool() {
+		return errors.New("no email has been confirmed as verified")
+	}
```
FreddyDevelop, line 21: When emailVerification is disabled and I want to verify my email in the profile I get an error when executing the actions `resend_passcode` or `back` or when I enter an invalid passcode e.g. `123456`
```
Code: technical_error, Message: Something went wrong., Cause: the action failed to handle the request: error while executing after hook actions: failed to execute hook action after state: no email has been confirmed as verified
```

FreddyDevelop (approved):

PR #1384, opened 2024-03-07
feat(flow-api): enable webauthn login with conditional mediation

lfleischmann:
# Description

Enables the flow API to handle WebAuthn/passkey logins with conditional mediation.

# Implementation

- Modifies the `send_capabilities` action by adding a new input for providing information about availability of conditional mediation on a client and stashing the value for further use.
- Adds and applies a login flow hook that generates WebAuthn request options before  the `login_init` state if conditional mediation is available (see point above) and applies the options to the response payload.
- Applies the existing action for verifying a WebAuthn assertion  to the `login_init` state. Because said action is now used on both mediated and un-mediated logins, it needs to know both the state the flow is in and the information about mediation availability in order to correctly suspend the action in the `login_init` state. Therefore I extended the `InitalizationContext` interface with a method to check whether the current state of a flow equals some other state. For this to work, the default implementation  of the `InitalizationContext` had to be extended with a field to hold the `FlowModel` in order to have access to state information.

# Tests

I have updated the `generic_client`  in the `backend/flow-api/static` directory so you should be able to test this with it.

# TODO

Because the `send_capabilities` check is used on all (non sub-)flows the mediation input is also present in flows where it is not relevant, i.e. registration and profile. Because `Flow`s / `FlowModel`s do not have any sort of "name" that generalizes over actual _instances_ of a flow (which means: instances have IDs but I cannot use that in this scenario because they differ from instance to instance ...) there is no way to know what flow an action/hook is actually executed in during initialization of an action (or during execution/finalization for that matter ...). This means there is currently no clear cut way to determine that the conditional mediation input in the preflight state should be omitted on registration and profile flows.

### DIFF
--- a/backend/flow_api/flow/login/hook_webauthn_generate_request_options_cond.go
+++ b/backend/flow_api/flow/login/hook_webauthn_generate_request_options_cond.go
@@ -34,11 +34,6 @@ func (a WebauthnGenerateRequestOptionsForConditionalUi) Execute(c flowpilot.Hook
 		return fmt.Errorf("failed to stash webauthn_session_data_id: %w", err)
 	}
 
-	err = c.Stash().Set("user_id", sessionDataModel.UserId)
-	if err != nil {
-		return fmt.Errorf("failed to stash user_id: %w", err)
-	}
-
 	err = c.Payload().Set("request_options", requestOptions)
 	if err != nil {
 		return fmt.Errorf("failed to set request_options payload: %w", err)


backend/flow_api/flow/login/hook_webauthn_generate_request_options_cond.go
```diff
@@ -0,0 +1,48 @@
...
+
+	err = c.Stash().Set("webauthn_session_data_id", sessionDataModel.ID)
+	if err != nil {
+		return fmt.Errorf("failed to stash webauthn_session_data_id: %w", err)
+	}
+
+	err = c.Stash().Set("user_id", sessionDataModel.UserId)
+	if err != nil {
+		return fmt.Errorf("failed to stash user_id: %w", err)
+	}
```
FreddyDevelop, line 40: I think we can remove this, because the userID will always be a nil uuid.

FreddyDevelop (approved):

Issue #1445, opened 2024-04-23
[FEAT]: Do not panic if getting SAML metadatdata fails


FreddyDevelop:
### Checklist

- [X] I could not find a solution in the existing issues or docs.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/teamhanko/hanko/blob/main/CODE_OF_CONDUCT.md).

### Description

Currently Hanko gets metadata for all SAML connections on startup and when a call to a metadata endpoint fails (e.g. http code > 400, wrong response body) Hanko won't start. 

### Describe your ideal solution

Hanko should show a warning in the logs when it fails to get SAML metadata, but should start anyway.

### Hanko Version

v0.10.2

shentschel:
So we want to ignore that provider after the warning?

FreddyDevelop:
> So we want to ignore that provider after the warning?

Yes

PR #1464, opened 2024-05-17
feat(saml): make getting providers from metadata non-panic

shentschel:
# Description

Changes the error behaviour when initialising saml providers. Instead of stoping the application now there will be a warning that the provider could not be loaded. The provider in question will be ignored on startup.

Closes: #1445


FreddyDevelop (commented):
Can we add information (e.g. domain or name) to the error message to identify the saml provider which is not working?
I got this error message while testing the PR, but when you have multiple identity provider it is hard/impossible to identify the provider which is not working: `failed to initialize provider: unable to unmarshal idp metadata response body to xml: expected element type <EntityDescriptor> but have <html>`

shentschel:
I extended the information with the idpconfig provider name.

FreddyDevelop (approved):

PR #1515, opened 2024-07-12
feat: add webhook env var decoder

lfleischmann:
# Description

The `webhook.hooks` config option is an array of objects. When using environment variables, specifying the hooks such that the `envconfig` package can correctly map the values into the config struct does not seem to be possible with the built-in capabilities.

# Implementation

`envconfig` allows defining custom decoders, though. This PR introduces an implementation so that hooks can be defined via environment variables as follows:

```shell
WEBHOOKS_HOOKS={"callback":"http://app.com/usercb","events":["user"]};{"callback":"http://app.com/emailcb","events":["email.send"]}
```

### DIFF
--- a/backend/config/webhook_config.go
+++ b/backend/config/webhook_config.go
@@ -1,9 +1,11 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


bjoern-m (approved):
LGTM

PR #1519, opened 2024-07-17
add flow api logging

FreddyDevelop:
# Description

Uses zerolog to log requests to the flow API. The logging middleware from echo could be used but an error from the flow API would not be logged. Thats because the handler function would need to return an error in order to be logged by the middleware, but the flow API handler does not return an error. So the handler logs the requests itself with zerolog.

# TODO

The logs from the flow API handler and from the middleware should log the same values. Currently the middleware logs more values that the flow API handler. We should check which values are needed and which we can remove.

### DIFF
--- a/backend/flow_api/handler.go
+++ b/backend/flow_api/handler.go
@@ -1,8 +1,11 @@
 package flow_api
 
 import (
+	"fmt"
 	"github.com/gobuffalo/pop/v6"
 	"github.com/labstack/echo/v4"
+	"github.com/rs/zerolog"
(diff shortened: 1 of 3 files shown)


bjoern-m:
Shall we remove the line `c.Logger().Errorf("failed to handle the request: %v", err)` in `backend/flow_api/handler.go: Line 90`?

FreddyDevelop:
Oh, yes we don't need that anymore.

bjoern-m (approved):

PR #1524, opened 2024-07-22
refactor: profile endpoint session validation

lfleischmann:
# Description

The sesssion middleware responses returned when invalid session tokens are provided has a different structure than other flow API error responses. We should try to make the responses consistent.

# Implementation

The "old" session middleware is no longer applied to the profile endpoint. A private method on the handler now validates the session. The method's implementation is basically a copy of the `echo-jwt` implementation (see [here](https://github.com/labstack/echo-jwt/blob/main/jwt.go#L201-L264)).


bjoern-m (commented):
Everything looks good and seems to be working; however, the translation for the new flow error is missing:
Screenshot 2024-07-22 at 13 09 25

bjoern-m (approved):

PR #1545, opened 2024-08-01
docs: flow api config docs and markdown generation

lfleischmann:
# Description

Document the configuration file and allow generation of markdown from jsonschema. 

# Implementation

- Provides a more complete documentation of config options using the https://github.com/invopop/jsonschema package (new config options, previously undocumented options like SAML and webhooks).
- Uses https://github.com/adobe/jsonschema2md to generate markdown files from the jsonschema so they can be used in the github wiki. 
    - There is a new command (schema json2md) that generates the files. The generated files can be found in `backend/.generated/config/md`.  The main "entrypoint" is the `hanko.md` file. The command deliberately does not use the existing schema generation "command" in `backend/json_schema` because for the output to have the desired structure, the jsonschema reflector must be configured differently. The command currently is not very flexible (Input/output paths are hardcoded, it generates docs for the config only but not the user import schema).
    - There is a new workflow that pushes the generated files to the wiki on manual workflow dispatch. I have only tested this in a private fork, should only be executed once we release.

### DIFF
--- a/backend/json_schema/hanko.config.json
+++ b/backend/json_schema/hanko.config.json
@@ -392,7 +392,7 @@
         },
         "optional": {
           "type": "boolean",
-          "description": "`optional` determines whether an email address must be provided when an input for this identifier is available.\nIf set to `true` this action can be skipped.",
+          "description": "`optional` determines whether users must provide an email when prompted.\nThere must always be at least one email address associated with an account. The primary email address cannot be\ndeleted if emails are required (`optional`: false`).",
           "default": true
         },
         "passcode_ttl": {
@@ -627,7 +627,7 @@
         },
         "optional": {
           "type": "boolean",
-          "description": "`optional` determines whether users must have registered at least one passkey. It controls whether passkey\ncreation can be skipped if prompted for. It also takes part in determining\nthe order of passkey and password acquisition on login and registration (see also `acquire_on_login` and\n`acquire_on_registration`).",
+          "description": "`optional` determines whether users must create a passkey when prompted. The last remaining passkey cannot be\ndeleted if passkeys are required (`optional: false`).\n\nIt also takes part in determining the order of password and passkey acquisition\non login and registration (see also `acquire_on_login` and `acquire_on_registration`): if one credential type is\nrequired (`optional: false`) then that one takes precedence, i.e. is acquired first.",
           "default": true
         },
         "user_verification": {
@@ -698,7 +698,7 @@
         },
         "optional": {
           "type": "boolean",
-          "description": "`optional` determines whether users must have a password set. It controls whether password creation can be\nskipped if prompted for. It also takes part in determining the order of password and passkey acquisition\non login and registration (see also `acquire_on_login` and `acquire_on_registration`).",
+          "description": "`optional` determines whether users must set a password when prompted. The password cannot be deleted if\npasswords are required (`optional: false`).\n\nIt also takes part in determining the order of password and passkey acquisition\non login and registration (see also `acquire_on_login` and `acquire_on_registration`): if one credential type is\nrequired (`optional: false`) then that one takes precedence, i.e. is acquired first.",
           "default": false
         },
         "recovery": {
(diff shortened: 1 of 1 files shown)


backend/config/config.go
```diff
@@ -519,16 +667,51 @@ func (r *RateLimiter) Validate() error {
...
+	// `providers` contains the configurations for the available OAuth/OIDC identity providers.
+	Providers ThirdPartyProviders `yaml:"providers" json:"providers,omitempty" koanf:"providers" jsonschema:"title=providers,uniqueItems=true"`
+	// `redirect_url` is the URL the third party provider redirects to with an authorization code. Must consist of the base URL
+	// of your running Hanko backend instance and the `callback` endpoint of the API,
+	// i.e. `{YOUR_BACKEND_INSTANCE}/thirdparty/callback.`
+	//
+	// Required if any of the [`providers`](#providers) are `enabled`.
+	RedirectURL string `yaml:"redirect_url" json:"redirect_url,omitempty" koanf:"redirect_url" split_words:"true" jsonschema:"example=https://yourinstance.com/thirdparty/callback"`
+	// `error_redirect_url` is the URL the backend redirects to if an error occurs during third party sign-in.
+	// Errors are provided  as 'error' and 'error_description' query params in the redirect location URL.
```
FreddyDevelop, line 686:
```suggestion
	// Errors are provided as 'error' and 'error_description' query params in the redirect location URL.
```

backend/config/config.go
```diff
@@ -581,11 +764,34 @@ func (t *ThirdParty) PostProcess() error {
 }
 
 type ThirdPartyProvider struct {
-	Enabled      bool   `yaml:"enabled" json:"enabled" koanf:"enabled"`
-	ClientID     string `yaml:"client_id" json:"client_id" koanf:"client_id" split_words:"true"`
-	Secret       string `yaml:"secret" json:"secret" koanf:"secret"`
-	AllowLinking bool   `yaml:"allow_linking" json:"allow_linking" koanf:"allow_linking" split_words:"true"`
-	DisplayName  string `jsonschema:"-" yaml:"-" json:"-" koanf:"-"`
+	// `allow_linking` indicates whether existing accounts can be automatically linked with this provider.
```
FreddyDevelop, line 767: Maybe we should document that linking is done through the email address.

backend/config/config.go
```diff
@@ -601,12 +807,18 @@ func (p *ThirdPartyProvider) Validate() error {
...
+	Apple ThirdPartyProvider `yaml:"apple" json:"apple,omitempty" koanf:"apple"`
+	// `discord` contains the provider configuration for Discord.
+	Discord ThirdPartyProvider `yaml:"discord" json:"discord,omitempty" koanf:"discord"`
+	// `github` contains the provider configuration for GitHub.
+	GitHub ThirdPartyProvider `yaml:"github" json:"github,omitempty" koanf:"github"`
+	// `google` contains the provider configuration for Google.
+	Google ThirdPartyProvider `yaml:"google" json:"google,omitempty" koanf:"google"`
+	// `linkedin` contains the provider configuration for LinkedIn.
+	LinkedIn ThirdPartyProvider `yaml:"linkedin" json:"linkedin,omitempty" koanf:"linkedin"`
+	// `microsoft` contains the provider configuration for Microsoft.
```
FreddyDevelop, line 820: The comments should be consistent: `... configuration for Sign in with Apple` - `... configuration for Discord.`

backend/config/config.go
```diff
@@ -479,17 +614,30 @@ var (
...
+	Store RateLimiterStoreType `yaml:"store" json:"store,omitempty" koanf:"store" jsonschema:"default=in_memory,enum=in_memory,enum=redis"`
+	// `redis_config` configures connection to a redis instance.
+	// Required if `store` is set to `redis`
+	Redis *RedisConfig `yaml:"redis_config" json:"redis_config,omitempty" koanf:"redis_config"`
+	// `passcode_limits` controls rate limits specific to the password/login endpoint.
+	PasscodeLimits RateLimits `yaml:"passcode_limits" json:"passcode_limits,omitempty" koanf:"passcode_limits" split_words:"true"`
+	// `password_limits` controls rate limits specific to the password/login endpoint.
+	PasswordLimits RateLimits `yaml:"password_limits" json:"password_limits,omitempty" koanf:"password_limits" split_words:"true"`
+	// `token_limits` controls rate limits specific to the token endpoint.
+	TokenLimits RateLimits `yaml:"token_limits" json:"token_limits,omitempty" koanf:"token_limits" split_words:"true" jsonschema:"default=token=3;interval=1m"`
```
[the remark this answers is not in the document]
FreddyDevelop: Yeah, I think that would work.

FreddyDevelop (approved):

PR #1625, opened 2024-09-10
Fix cookie name in passcode handler

### LINKED ISSUE #1624: Self-Hosted hanko with session cookie name does not allow to add additional email addresses
### Checklist

### Describe the bug

Setup self-hosted hanko. Change the session cookie name in your config.

Try to add an email address to an existing login fails.

### Reproducing the bug

1. run self-hosted hanko
2. login to an existing account
3. add additional email on profile
4. boooom 💥

### Logs

```Shell
code=403, message=Forbidden, internal=passcode finalization not allowed
```

### Configuration

```yml
session:
  cookie:
    name: "foo"
```

### Hanko Version

v1.0.2

### OS Hanko Backend

None

### OS Version Hanko Backend

_No response_

### OS

None

### OS Version

_No response_

### Browser Version

_No response_

### Environment

Other (e.g. Kubernetes)

### Additional Context

_No response_

loeffert:
# Description

Use configured cookie name in passcode handler. Fixes #1624

### DIFF
--- a/backend/handler/passcode.go
+++ b/backend/handler/passcode.go
@@ -3,13 +3,17 @@ package handler
 import (
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
 	"github.com/gobuffalo/pop/v6"
 	"github.com/gofrs/uuid"
 	"github.com/labstack/echo/v4"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	zeroLogger "github.com/rs/zerolog/log"
 	"github.com/sethvargo/go-limiter"
-	"github.com/teamhanko/hanko/backend/audit_log"
+	auditlog "github.com/teamhanko/hanko/backend/audit_log"
 	"github.com/teamhanko/hanko/backend/config"
 	"github.com/teamhanko/hanko/backend/crypto"
 	"github.com/teamhanko/hanko/backend/dto"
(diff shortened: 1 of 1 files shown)


backend/handler/passcode.go
```diff
@@ -453,7 +452,7 @@ func (h *PasscodeHandler) Finish(c echo.Context) error {
 
 func (h *PasscodeHandler) GetSessionToken(c echo.Context) jwt.Token {
 	var token jwt.Token
-	sessionCookie, _ := c.Cookie("hanko")
+	sessionCookie, _ := c.Cookie(h.cfg.Session.Cookie.GetName())
```
loeffert, line 455: @FreddyDevelop how can we handle the errors here? all errors in GetSessionToken are ignored. They should be logged at least. What is the best way to get a logger instance here?

That would have helped to faster find that bug, because the accessed cookie doesn't exist.

FreddyDevelop: We do not log the error, that a cookie is missing, because the session token can also be send within the `Authorization` header and when the error gets logged then we would clutter the logs with unnecessary messages. 

Only solution to that would be to log that errors only when no token can be returned. You can use zerolog for the logging. An example can be found at: https://github.com/teamhanko/hanko/blob/579cef31a136c6cf3cb650d6e885f9595c8cda2c/backend/handler/passcode.go#L234

bjoern-m (approved):
Approving this as it resolves an issue. Error logging can be addressed in a separate PR.

PR #1670, opened 2024-10-07
Remove Handler for Account Enumeration Attack

EasterTheBunny:
# Description

OWASP ID: WSTG-IDNT-04

Reference URL:
https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account

Account enumeration is enabled by the handler `/user` where a user email can be provided and the server provides a validation response of whether the user email exists along with other data.

# Implementation

- The server handler for the endpoint `/user` was removed
- The frontend-sdk function `getInfo` which accepts an email was removed

# Tests

Tests were removed that validated previous functionality.

### DIFF
--- a/backend/handler/public_router.go
+++ b/backend/handler/public_router.go
@@ -2,11 +2,12 @@ package handler
 
 import (
 	"fmt"
+
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/sethvargo/go-limiter"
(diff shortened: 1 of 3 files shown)


FlxMgdnz:
Thanks @EasterTheBunny. We're aware. Due to backwards compatibility requirements we cannot merge this PR at the moment, but we will when the time is right.

PR #1736, opened 2024-10-14
create a hanko post on dev.to

DanishAliUmar:
# Description

create a hanko post on dev.to

# Implementation
Wrote an article which can be found it the link below:
https://dev.to/danishali/embracing-passwordless-authentication-passkeys-and-hanko-abl

image
# Todos

Dev.to


FlxMgdnz:
/award 750 Thanks! However, I gotta say that the level of the article is not very high and it also looks a lot like AI to me. Still -- it's good to get the word out.

PR #1899, opened 2024-10-26
feat: hanko bug finding tasks

adilkadivala:
# Description

Bug hunting

### Points

50 - 150 

### Description

only loads a component which is clicked

in login and create accont page 

when click on signup with google or github

then both are loads

### Provide proof that you've completed the task

hancobug2

hancobug


FlxMgdnz:
Can you explain what you think the bug is? I don't understand the report, sorry.

adilkadivala:
when I click on any one button then loading both ;;

If i click on google then loading google and github both, and same when I click on github

adilkadivala:
hey @FlxMgdnz please check out my all side-quest ,  I have completes 6 ,  but just got 150 points

adilkadivala:
my all side-quest [here](https://github.com/teamhanko/hanko/issues/created_by/@me)

PR #1957, opened 2024-11-04
feat: add last used indicator to the login page

bjoern-m:
# Description

Adds a last-used indicator to the passkey and third-party buttons.


FlxMgdnz:
@bjoern-m I've looked at the PR and would suggest that we only add the Last Used indicator to the social buttons for now. Adding this information to the other options could be unfamiliar for the user and also lead to misunderstandings.

PR #1969, opened 2024-11-07
feat: add session creation endpoint

FreddyDevelop:
# Description

Add a session creation endpoint in the admin API.


bjoern-m:
Shall we add a config parameter to enable or disable the new endpoint for security reasons?

FreddyDevelop:
> Shall we add a config parameter to enable or disable the new endpoint for security reasons?

I don't think that we need a config parameter for this, because when an attacker wants to have control over an account and has access to the admin API he can just add his authentication factor (currently he can just add an email address) to the user and he can login and can remove the email address from the profile or even from the admin API. So the session endpoint only makes it easier for an attacker.

PR #1978, opened 2024-11-20
Feat custom user

FreddyDevelop:
# Description

Add custom user handle to a webauthn credential. This custom user handle is only used for verifying assertions from imported passkeys.
New passkeys are created with the user id (uuid) from Hanko.

# Implementation

Added a new column to the webauthn credentials table. Check in `WebauthnService.VerifyAssertionResponse` if the userHandle in the response is a uuid, if not treat it as a custom user handle. Get the user based on the credential ID verify the response against the user credentials

# Tests

- Create a passkey for a user with an user which is not a uuid. (E.g. with https://github.com/teamhanko/passkeys)
- Copy the webauthn credential into the database (currently there is no other way to import a webauthn credential)
- Add the user handle to the newly created webauthn credential
- Try to login using this new credential

Keep in mind that the webauthn credential must be created and used with the same RP ID.

# Todos

- Extend the user import to also import a custom user id

# Additional context

This is needed when a relying party already has implemented passkeys, but wants to migrate to Hanko and does not use uuids as user identifiers.

### DIFF
--- a/backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
+++ b/backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
@@ -1,0 +1,1 @@
+add_column("webauthn_credentials", "custom_user_handle", "string", { "null": true })
--- a/backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
+++ b/backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
@@ -13,4 +13,4 @@ add_foreign_key("webauthn_credentials", "user_handle_id", {"webauthn_credential_
     "on_update": "cascade",
 })
 
-sql("ALTER TABLE webauthn_credentials ADD CONSTRAINT webauthn_credential_user_handle_fkey FOREIGN KEY (user_handle_id, user_id) REFERENCES webauthn_credential_user_handles(id, user_id) ON DELETE ON UPDATE CASCADE;")
+sql("ALTER TABLE webauthn_credentials ADD CONSTRAINT webauthn_credential_user_handle_fkey FOREIGN KEY (user_handle_id, user_id) REFERENCES webauthn_credential_user_handles(id, user_id) ON DELETE CASCADE ON UPDATE CASCADE;")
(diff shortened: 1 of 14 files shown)


backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
```diff
@@ -0,0 +1,16 @@
...
+    t.ForeignKey("user_id", {"users": ["id"]}, {"on_delete": "cascade", "on_update": "cascade"})
+}
+
+add_column("webauthn_credentials", "user_handle_id", "uuid", { "null": true })
+add_foreign_key("webauthn_credentials", "user_handle_id", {"webauthn_credential_user_handles": ["id"]}, {
+    "on_delete": "cascade",
+    "on_update": "cascade",
+})
+
+sql("ALTER TABLE webauthn_credentials ADD CONSTRAINT webauthn_credential_user_handle_fkey FOREIGN KEY (user_handle_id, user_id) REFERENCES webauthn_credential_user_handles(id, user_id) ON DELETE CASCADE ON UPDATE CASCADE;")
```
bjoern-m, line 16:
```suggestion
sql("ALTER TABLE webauthn_credentials ADD CONSTRAINT webauthn_credential_user_handle_fkey FOREIGN KEY (user_handle_id, user_id) REFERENCES webauthn_credential_user_handles(id, user_id) ON DELETE NO ACTION ON UPDATE CASCADE;")
```

backend/persistence/migrations/20241118114500_change_webauthn_credentials.up.fizz
```diff
@@ -0,0 +1,16 @@
...
+    t.Column("user_id", "uuid", {"null": false})
+    t.Column("handle", "string", {"null": false, "unique": true})
+    t.Timestamps()
+    t.Index(["id", "user_id"], {"unique": true})
+    t.ForeignKey("user_id", {"users": ["id"]}, {"on_delete": "cascade", "on_update": "cascade"})
+}
+
+add_column("webauthn_credentials", "user_handle_id", "uuid", { "null": true })
+add_foreign_key("webauthn_credentials", "user_handle_id", {"webauthn_credential_user_handles": ["id"]}, {
+    "on_delete": "cascade",
```
bjoern-m, line 12: I would tend to set the `user_handle_id` to `NULL` when the referenced entry in `webauthn_credential_user_handles` is deleted, as assigning a user handle is optional. Conversely, I think it makes sense to drop the user handle when the credential itself is deleted.
```suggestion
    "on_delete": "set null",
```

bjoern-m (approved):

PR #1992, opened 2024-12-05
Feat extend user import

FreddyDevelop:
# Description

Extend the user import to import existing passwords, webauthn credentials, ...

# Implementation

Extended the already existing import command to include more information.

# Tests

Create a json file with users to import, then use the import command to import the users.


bjoern-m (commented):
I tried to import the following file, that contains a user having two WebAuthn credentials with the same user handle. I think this should be working, but the importer tries to insert the user handle twice causing a unique key constraint violation.

[users.json](https://github.com/user-attachments/files/18062166/users.json)

bjoern-m (approved):

PR #2113, opened 2025-04-04
feat: custom jwt claims

lfleischmann:
# Description

Add the ability do define custom claims in the session token.

# Implementation

- Add a new configuration option `session.jwt_template` that allows defining `session.jwt_template.claims`. 
Values can be arbitrarily nested and can use Go text templates.
- The custom claims are also added to `sessions/validate` endpoint responses.
- Modified the `jwt create` command to pretty print the payload alongside the token. 

# Test

- Some unit tests for the new code are included.
- For manual testing:
	- configure claims in the config file then use the `jwt create` command to see if the claims are added to the token
	- use the generated token with the `sessions/validate` endpoints to see if the claims are also part of the response

### DIFF
--- a/frontend/frontend-sdk/README.md
+++ b/frontend/frontend-sdk/README.md
@@ -229,10 +229,13 @@ To allow for IDE autocompletion and to maintain type safety for your custom clai
 4. Add your custom claims to your custom type.
 
 ```ts
-import type { Claims } from "@teamhanko/hanko-elements" // 2.
+import type { Claims } from "@teamhanko/hanko-frontend-sdk" // 2.
+// import type { Claims } from "@teamhanko/elements"        // alternatively, if you use Hanko Elements, which
+                                                            // re-exports most SDK types
 
-type CustomClaims = Claims<{                            // 3.
-    custom_claim?: string                               // 4.
+
+type CustomClaims = Claims<{                                // 3.
+    custom_claim?: string                                   // 4.
 }>;
 ```
 


backend/config/config_session.go
```diff
@@ -97,3 +101,7 @@ type ServerSide struct {
 	// older sessions are invalidated.
 	Limit int `yaml:"limit" json:"limit,omitempty" koanf:"limit" jsonschema:"default=100"`
 }
+
```
bjoern-m, line 104: We could add some docs here, like:

```suggestion
// JWTTemplate defines a template for adding custom claims to session JWTs.
// These claims are processed at JWT generation time and can include static values,
// templated strings using Go's text/template syntax, or nested structures (maps and slices).
// The template has access to user data via the `.User` field, which includes:
// - `.User.UserID`: The user's unique ID (string)
// - `.User.Email`: Email details (optional, with `.Address`, `.IsPrimary`, `.IsVerified`)
// - `.User.Username`: The user's username (string, optional)
//
// Claims that fail to process (e.g., due to invalid templates) are logged and skipped,
// ensuring JWT generation continues without interruption.
//
// Example usage in YAML configuration:
// ```yaml
// session:
//   lifespan: 24h
//   jwt_template:
//     claims:
//       role: "user"                           # Static value
//       user_email: "{{.User.Email.Address}}"  # Templated string
//       is_verified: "{{.User.Email.IsVerified}}"  # Boolean from user data
//       metadata:                              # Nested map
//         source: "hanko"
//         greeting: "Hello {{.User.Username}}"
//       scopes:                                # Slice with templated value
//         - "read"
//         - "write"
//         - "{{if .User.Email.IsVerified}}admin{{else}}basic{{end}}"
// ```
//
// In this example:
// - `role` is a static string ("user").
// - `user_email` dynamically inserts the user's email address.
// - `is_verified` inserts a boolean indicating email verification status.
// - `metadata` is a nested map with a static `source` and a templated `greeting`.
// - `scopes` is a slice combining static values and a conditional template.
//
// Notes:
// - Templates must be valid Go `text/template` syntax. Invalid templates are logged and ignored.
// - Boolean strings ("true" or "false") from templates are automatically converted to actual booleans.
// - Use conditionals (e.g., `{{if .User.Email}}`) to handle optional fields safely.
//
// For more details on template syntax, see: https://pkg.go.dev/text/template
```

frontend/frontend-sdk/src/lib/Dto.ts
```diff
@@ -268,6 +268,7 @@ export interface Claims {
   email?: Pick<Email, "address" | "is_primary" | "is_verified">;
   username?: string;
   session_id: string;
+  [key: string]: unknown;
```
bjoern-m, line 271: Do you see a way to improve this one, so that the user can provide a type for the custom claims?

bjoern-m (approved):

bjoern-m (approved):

PR #2157, opened 2025-05-14
feat: user metadata

lfleischmann:
# Description

This PR introduces customizable user metadata. The metadata can be accessed in session JWT templates to map metadata to claims in a session token.

# Implementation

- Adds DB table `user_metadata` and corresponding `UserMetadata` model and persister
- Adds new Admin API metadata handler with endpoints for retrieving and patching metadata
   - Patching metadata uses [`evanphx/json-patch`](https://github.com/evanphx/json-patch), which in turn is based on [JSON merge patch](https://datatracker.ietf.org/doc/html/rfc7386)
- Makes the metadata available on the `User` context object in JWT templates
   - Querying of metadata uses [GJSON](https://github.com/tidwall/gjson) so JWT templates must use [GJSON path syntax](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) (see the tests for examples)

# Tests

- Adds tests for the new metadata handler
- `session/template_test.go`:
    - Extends tests for the central public template processing function
    - Removes tests for private template processing functions (they basically tested the same thing)

# Additional context

https://github.com/teamhanko/docs/pull/134 - guide(s) and API spec updates

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -494,7 +494,9 @@ three types that differ as to how they can be accessed and modified:
 | Public        | Read access                  | Read and write access |
 | Unsafe        | Read access and write access | Read and write access |
 
-Each metadata type is currently limited to 3000 characters.
+Each metadata type supports a maximum of 3,000 characters. Metadata is stored as compact JSON (whitespace is ignored).
+JSON syntax characters (`{`, `:`, `"`, `}`) count toward the character limit.
+Multibyte UTF-8 characters (like emojis or non-Latin characters) count as 1 character each.
 
 #### Private metadata
 


backend/handler/metadata_admin.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+		return echo.NewHTTPError(http.StatusInternalServerError, "could not save metadata").
+			SetInternal(err)
+	}
+
+	response := admin.NewMetadata(currentMetadataModel)
+
+	if response == nil {
+		return c.JSON(http.StatusNoContent, nil)
+	}
```
FreddyDevelop, line 101:
```suggestion
	if response == nil {
		return c.NoContent(http.StatusNoContent)
	}
```

FreddyDevelop (approved):

FreddyDevelop (approved):

PR #2197, opened 2025-07-22
feat: support nonce for styles

### LINKED ISSUE #1983: elements: support Content-Security-Policy
### Checklist

### Description

When using the default inline styles for elements, a Content-Security-Policy needs to have "unsafe-inline" allowed.

Hanko could have an option to accept a `nonce` to include with the inline styles, I think.

_No response_

### Hanko Version

latest

### Additional Context

I use `https://cdn.jsdelivr.net/npm/@teamhanko/hanko-elements/dist/elements.js` which I think is the latest version?

FreddyDevelop:
# Description

This PR adds support for the nonce property to allow a more stricter content-security-policy (no need for `unsafe-inline`) when using Hanko. 

Fixes #1983 

# Implementation

The components forward the nonce property from one of the elements (`hanko-auth`, `hanko-login`, `hanko-registration`) to the inline style tag. 

Unfortunately the nonce property is hidden for security reasons ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/nonce)) and can only be accessed through the IDL property (`script['nonce']`). That means that it cannot be accessed through props, e.g. we do with the `lang` property, but has to be done through `document.getElementsByTagName()[0].nonce`. This approach has one disadvantage, when using more than one of the same element on the same page, and both have different nonce values only one value will be used, but that should not break the elements.

# Tests

1. Add a CSP which disallows `unsafe-inline` for styles.
2. Add a nonce to the CSP for `style-src`.
3. Add the nonce to the hanko components (`hanko-auth`, ...)

The inline styles should be applied.

[src.zip](https://github.com/user-attachments/files/21363008/src.zip) contains changed files from the quickstart which can be used to test it.

# Todos

- Add a guide to the docs.

### DIFF
--- a/frontend/elements/src/Elements.tsx
+++ b/frontend/elements/src/Elements.tsx
@@ -84,17 +84,24 @@ const globalOptions: GlobalOptions = {};
 const createHankoComponent = (
   componentName: ComponentName,
   props: Record<string, any>,
-) => (
-  <AppProvider
-    componentName={componentName}
-    globalOptions={globalOptions}
-    createWebauthnAbortSignal={createWebauthnAbortSignal}
-    {...props}
-  />
-);
+) => {
+  props.nonce = document.getElementsByTagName(
+    `hanko-${componentName}`,
+    // @ts-ignore
+  )[0].nonce;
+  return (
+    <AppProvider
+      componentName={componentName}
+      globalOptions={globalOptions}
+      createWebauthnAbortSignal={createWebauthnAbortSignal}
+      {...props}
+    />
+  );
+};
 
-const HankoAuth = (props: HankoAuthElementProps) =>
-  createHankoComponent("auth", props);
+const HankoAuth = (props: HankoAuthElementProps) => {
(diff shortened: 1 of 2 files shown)


bjoern-m (commented):
I left a few suggestions, but it's looking good so far and the attached example works fine. Also, when I provide an invalid `nonce`, I can see that the browser blocks the content as intended.

image

image

bjoern-m:
> This approach has one disadvantage, when using more than one of the same element on the same page, and both have different nonce values only one value will be used, but that should not break the elements.

With this limitation in mind, would it be an option to set the `nonce` through the `register()` options instead?

FreddyDevelop:
> With this limitation in mind, would it be an option to set the `nonce` through the `register()` options instead?

Technically we can do that, but I wanted to match the implementation as close to the HTML standard as possible, to make it more intuitive to use.

bjoern-m:
The README needs an update to document the introduced attribute for the web components.

PR #2285, opened 2025-10-24
feat: add flow locking mechanism

lfleischmann:
# Description

Implements a configurable flow locking mechanism to prevent concurrent execution of the same flow, addressing possible race conditions and side effects in actions (setting header tokens, sending emails, ...) in concurrent request scenarios.

## Problem 
When multiple parallel requests attempt to process the same flow simultaneously (e.g., 5 concurrent passcode verification requests), race conditions occur leading to:
- Version conflicts on flow state updates - which is caught by our check on the flow version in the flowpilot, but only AFTER actions and hooks have run and the flow gets updated, so any side effect produced within actions will still take effect (like setting the `X-Auth-Token` header with a session token - which is not valid, but still ...)
- Wasted database operations (only one succeeds, others fail)

# Implementation
Introduces a `FlowLocker` interface with three implementations:

### 1. **In-Memory Locker** (default)
- Fail-fast locking using Go mutexes
- Suitable for single-instance deployments
- No external dependencies

### 2. **Redis Locker**
- Distributed locking using Redsync ([Redlock](https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/) algorithm - using the [`go-redsync`](github.com/go-redsync/redsync/v4) package)
- Suitable for multi-instance deployments
- Shared state across application instances

### 3. **No-Op Locker**
- Pass-through implementation when locking is disabled

### Key Features
- **Fail-fast behavior**: Second and consequent concurrent requests immediately fail rather than waiting
- **Context-aware**: Respects context cancellation
- **Configurable**: Can be enabled/disabled via configuration
- **Automatic cleanup**: Locks are automatically released via `unlock()` function returned from locking function

### Integration
The flow locker is integrated into the flow execution handler. Locking takes effect for all flow executions except for those that initialize a flow. 

## Configuration

```yaml
flow_locker:
  enabled: true  # default: true
  store: in_memory  # options: in_memory, redis
  ttl: 30s  # lock expiry (Redis only)
  redis_config:  # required if store is redis
    address: "localhost:6379"
    password: "secret"
```

# Tests
Test coverage includes:
- ✅ Unit tests for all locker implementations
- ✅ Concurrent access tests (10-100 goroutines)
- ✅ Race condition detection (run with `-race` flag)
- ✅ Context cancellation behavior
- ✅ Lock expiry (Redis)
- ✅ Multiple instance coordination (Redis)
- ✅ Table-driven tests for factory function
- ✅ Dockertest integration for Redis tests


backend/flow_api/flow_locker/flow_locker.go
```diff
@@ -0,0 +1,39 @@
...
+func NewFlowLocker(cfg config.FlowLocker) (FlowLocker, error) {
+	if !cfg.Enabled {
+		return NewNoOpLocker(), nil
+	}
+
+	switch cfg.Store {
+	case config.FLOW_LOCKER_STORE_REDIS:
+		if cfg.Redis == nil {
+			return nil, fmt.Errorf("redis config required for redis flow locker")
+		}
```
FreddyDevelop: I think this check can be removed, because the check is already done in the config validation.

FreddyDevelop (approved):

PR #2336, opened 2025-12-11
fix: user verification on passkey creation

oskimora13:
The UV value was hardcoded, but is now read from the configuration.

# Description

# Implementation

# Tests

# Todos

# Additional context


FlxMgdnz:
Was fixed a while ago with #2239

oskimora13 (commented):
Yes

PR #2343, opened 2025-12-16
feat: make oauth client secret config optional

MarcoScabbiolo:
When setting up a Google OIDC client using Android as the client type Google does not generate a client secret. Hanko does not boot up if its not provided a client secret, and providing a wrong client secret makes Google reject the request. Making the client secret optional allows the underlying oAuth library to do the PKCE flow correctly.


FlxMgdnz:
Hey @MarcoScabbiolo, thanks for your contribution. Could you describe what exactly you are planning to do, or in which scenario you want to use Hanko? Our 3rd-party flows currently still rely on providers like Google redirecting back to us, meaning to the Hanko backend, which means we essentially always require the client secret. We do have native app flows on our radar, but so far we have focused more on the web use case during development.

MarcoScabbiolo:
@FlxMgdnz Thanks for the quick response! I am implementing a POC of Hanko in Android and the usual Android OIDC flow redirects back to the app instead of the backend, at least for the Google provider. 

The way I have it working (by making the client secret optional) is  the app handling the OIDC callback from google and calls Hanko's BE `/thirdparty/callback` itself to retrieve the token, and after that performing the `exchange_token` action with the token. 

Google's OIDC is configured with the app's package name as the redirect link, google redirects back to the app using an Android App Link.

> We do have native app flows on our radar, but so far we have focused more on the web use case during development.

I noticed, but the gap to support native apps (other than making the client secxret optional) is mainly in data format. For example sending data via cookies which is tedious to handle outside of web contexts, but not impossible.

There are other flows that have more severe gaps but I didn't dive deep into them so I might be missing something:
- MFA has no recovery mechanism. Even if setting it up using the device's biometric to encrypt the TOTP secret is possbile, there is no way to recover it later making it only usable with admin intervention.
- Passkeys I haven't even looked at TBH, but it looks like there is no straightforward way to set it up outside of webauthn.

Of course using the web version of Hanko and a WebView or even opening the browser is an option, but I am trying to see how far I can get with a purely native setup before falling back to the web only version.

MarcoScabbiolo:
I can share my POC repo privately and I am open to delegate its ownership later if it gets to a good enough point, but I'm sure you're well aware the costs of polishing and making a codebase open source ready. This is a side/personal project for me, not a full time investment.

PR #2356, opened 2026-01-07
feat(i18n): add dutch translations (backend mailing)

Harm-Nullix:
# Description

Added dutch translations.
I think it is a good addition to the default sets.

Been using Hanko since everything is slightly moving towards European providers so I think it is a wise choice to add dutch.

# Implementation

Followed all places that hold "fr" and "it" to find places where the keys are used.
Also checked the initial commits for the fr.ts and it.ts file to see if anything is missing, but it seems more complete than those.

Follow-up on #2352, where that was frontend based, this is based on adding the translation file for the mailing.
(I did not see any other files than just the passcode.yaml file.


lfleischmann (approved):
LGTM, thanks!

PR #2402, opened 2026-01-29
fix: use 400 Bad Request for expired passcode instead of 408

### LINKED ISSUE #2398: Incorrect HTTP status code for expired passcode (408 instead of 400/422)
## Description

Found in `backend/handler/passcode.go` around line 319: When a passcode expires, the API returns `http.StatusRequestTimeout` (408), which is semantically incorrect for this use case.

## Current Code

```go
businessError = echo.NewHTTPError(http.StatusRequestTimeout, "passcode request timed out").SetInternal(errors.New(fmt.Sprintf("createdAt: %s -> lastVerificationTime: %s", passcode.CreatedAt, lastVerificationTime))) // TODO: maybe we should use BadRequest, because RequestTimeout might be to technical and can refer to different error
```

## Problem

**HTTP 408 Request Timeout** is meant for situations where:
- The client took too long to send the request
- The server timed out waiting for the request body
- Network/connection issues caused a timeout

**It should NOT be used when:**  
- Application-level validation fails (like an expired passcode)
- Business logic determines something is no longer valid

## Impact

1. **Client confusion**: HTTP clients may retry 408 responses automatically, which is inappropriate for expired passcodes
2. **Monitoring issues**: 408s typically indicate infrastructure problems, not application-level validation failures
3. **API semantics**: Violates HTTP specification intent

## Recommended Fix

Use a more appropriate status code:

### Option 1: `400 Bad Request` (Recommended)
```go
businessError = echo.NewHTTPError(http.StatusBadRequest, "passcode has expired").SetInternal(errors.New(fmt.Sprintf("createdAt: %s -> lastVerificationTime: %s", passcode.CreatedAt, lastVerificationTime)))
```

### Option 2: `422 Unprocessable Entity`
```go
businessError = echo.NewHTTPError(http.StatusUnprocessableEntity, "passcode has expired").SetInternal(errors.New(fmt.Sprintf("createdAt: %s -> lastVerificationTime: %s", passcode.CreatedAt, lastVerificationTime)))
```

## Location

`backend/handler/passcode.go`, line ~319 (in the `Finish` method)

## References

- [RFC 7231 - HTTP Status Codes](https://tools.ietf.org/html/rfc7231#section-6.5.7)
- [MDN - 408 Request Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408)

jaffarkeikei:
Resolves #2398

## Summary
Changed HTTP status code from 408 (Request Timeout) to 400 (Bad Request) for expired passcodes.

## Changes
- Status: `http.StatusRequestTimeout` → `http.StatusBadRequest`
- Message: "passcode request timed out" → "passcode has expired"
- Removed TODO comment

## Why?
HTTP 408 is for network/client timeouts, not application validation failures. Using 400 prevents clients from auto-retrying and correctly indicates a client error.

## Impact
- ✅ Better HTTP semantics (aligns with RFC 7231)
- ✅ No automatic retries from HTTP clients
- ✅ Clearer error handling

### DIFF
--- a/backend/handler/passcode.go
+++ b/backend/handler/passcode.go
@@ -309,7 +309,7 @@ func (h *PasscodeHandler) Finish(c echo.Context) error {
 			if err != nil {
 				return fmt.Errorf("failed to create audit log: %w", err)
 			}
-			businessError = echo.NewHTTPError(http.StatusRequestTimeout, "passcode request timed out").SetInternal(errors.New(fmt.Sprintf("createdAt: %s -> lastVerificationTime: %s", passcode.CreatedAt, lastVerificationTime))) // TODO: maybe we should use BadRequest, because RequestTimeout might be to technical and can refer to different error
+			businessError = echo.NewHTTPError(http.StatusBadRequest, "passcode has expired").SetInternal(errors.New(fmt.Sprintf("createdAt: %s -> lastVerificationTime: %s", passcode.CreatedAt, lastVerificationTime)))
 			return nil
 		}
 


lfleischmann:
Won't do. Old API endpoints are already marked deprecated and slated for removal in the near future. Closing this PR.

PR #2429, opened 2026-02-19
feat: add AMR values to session token

lfleischmann:
# Description

Add `amr` values to session token claims and session validation response.

# Implementation

- Derive login AMR values from already present `login_method/mfa_method/third_party_provider` stash values in a hook before session issuance for login + token exchange flows.
- Track registration “used/enrolled” factors (passcode + created credentials) via dedicated stash keys and derive `amr` values in a hook before issuing the session.
- Set the derived `amr` values in issue session hook when generating/creating the session as additional JWT option.


backend/flow_api/flow/shared/const_flow_names.go
```diff
@@ -7,6 +7,7 @@ const (
 	FlowCredentialOnboarding flowpilot.FlowName = "credential_onboarding"
 	FlowCredentialUsage      flowpilot.FlowName = "credential_usage"
 	FlowDeviceTrust          flowpilot.FlowName = "device_trust"
+	FlowExchangeToken        flowpilot.FlowName = "exchange_token"
```
FreddyDevelop: The variable should also be used at Flow creation.

PR #2641, opened 2026-05-24
Add webhook events for session creation and deletion

### LINKED ISSUE #2631: Add webhook events for session creation and deletion
### Checklist

### Description

**Problem**
When using Hanko in a hybrid setup where JWTs are validated locally (e.g., by extracting Hanko's JWKS and verifying tokens on your own backend), there is no way to be notified when a session is revoked or deleted.

Hanko stores sessions server-side and allows users to revoke them via the profile UI (when session.allow_revocation is enabled), but the JWT itself remains cryptographically valid until its exp claim passes. For sure this is by design.

**Proposal**
Add the following webhook events:

session.created — fired when a new session is issued
session.deleted — fired when a session is revoked by a user, deleted by an admin, or invalidated due to idle timeout / session limit enforcement
The webhook payload should include at minimum:

session_id
user_id
JWT exp timestamp (so consumers know when the blacklist entry can be evicted)

Call /sessions/validate on every request — defeats the purpose of local JWT validation and introduces latency + availability dependency on Hanko.

Use short-lived JWTs — reduces the revocation gap but doesn't eliminate it, and there's no silent token refresh mechanism to make this seamless for users.

Poll Hanko's session list API — works but is inefficient and still has a delay window equal to the polling interval.

### Hanko Version

cloud

### Additional Context

_No response_

samadrehman:
This PR introduces new webhook events to provide real-time notifications for session lifecycle changes, specifically session creation and deletion. This addresses a critical need for hybrid Hanko setups where JSON Web Tokens (JWTs) are validated locally, but server-side session revocations or expirations are not immediately propagated to the backend.

Problem Addressed: In a hybrid Hanko environment, JWTs remain cryptographically valid until their exp claim passes, even if the session is revoked or deleted server-side (e.g., by a user via the profile UI, by an admin, or due to idle timeout/session limit enforcement). This can lead to a "revocation gap" where a locally validated JWT grants access to resources even after the corresponding session has been terminated on the Hanko backend. Existing workarounds like calling /sessions/validate on every request introduce latency and availability dependencies, while short-lived JWTs only reduce the gap without eliminating it.

Ideal Solution Implemented: To enable immediate notification of session state changes, the following webhook events have been added:

session.create: Fired when a new session is issued.
session.delete: Fired when a session is revoked by a user, deleted by an admin, or invalidated due to idle timeout/session limit enforcement.
Webhook Payload: The payload for these webhook events will include at minimum:

session_id: The unique identifier of the session.
user_id: The ID of the user associated with the session.
exp timestamp: The expiration timestamp of the JWT, allowing consumers to know when a blacklist entry can be evicted.
Changes Made:

[d:\hanko\backend\json_schema\hanko.config.json](code-assist-path:d:\hanko\backend\json_schema\hanko.config.json): The JSON schema has been updated to include session, session.create, and session.delete as valid webhook event types.
[d:\hanko\backend\README.md](code-assist-path:d:\hanko\backend\README.md): The webhook event types table in the documentation has been updated to reflect the new session events and their triggers.

close  #2631

### DIFF
--- a/backend/config/config_test.go
+++ b/backend/config/config_test.go
@@ -9,6 +9,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMain(m *testing.M) {
+	_ = os.Setenv("DEBUG", "false")
+	os.Exit(m.Run())
+}
+
 func TestDefaultConfigAccountParameters(t *testing.T) {
 	cfg := DefaultConfig()
 	assert.Equal(t, cfg.Account.AllowDeletion, false)
--- a/backend/handler/session.go
+++ b/backend/handler/session.go
@@ -10,6 +10,7 @@ import (
 	"github.com/teamhanko/hanko/backend/v2/config"
 	"github.com/teamhanko/hanko/backend/v2/dto"
 	"github.com/teamhanko/hanko/backend/v2/persistence"
+	webhookutils "github.com/teamhanko/hanko/backend/v2/webhooks/utils"
 	"github.com/teamhanko/hanko/backend/v2/session"
 )
 
@@ -66,6 +67,9 @@ func (h *SessionHandler) ValidateSession(c echo.Context) error {
 				if sessionDeletionErr != nil {
 					return fmt.Errorf("failed to delete session: %w", sessionDeletionErr)
 				}
+
+				// notify webhook about deleted session
+				webhookutils.NotifySessionDelete(c, nil, *sessionModel)
 
 				cookie, cookieDeletionErr := h.sessionManager.DeleteCookie()
 				if cookieDeletionErr != nil {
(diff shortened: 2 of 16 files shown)


FlxMgdnz:
Hi @samadrehman,

is it possible that this PR is incomplete? Just aksing, because the changes mostly include Schema/README changes but they do not change the backend’s behaviour in terms of triggering actual Webhooks on occurence of the events described in the original issue. If you deem it as complete: I am sorry to say that simply updating documentation artifacts is not sufficient for this change.

Let me know wht your current status on this is.

Thanks

samadrehman:
Sorry @FlxMgdnz  , I not knew why this changes only happened currently checking it and update in few minutes . Sorry for any inconvenience occurred .

backend/config/config_test.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMain(m *testing.M) {
+	_ = os.Setenv("DEBUG", "false")
+	os.Exit(m.Run())
+}
+
```
lfleischmann: What is this for?

backend/handler/session.go
```diff
@@ -133,21 +137,24 @@ func (h *SessionHandler) ValidateSessionFromBody(c echo.Context) error {
 
 	// Check idle timeout
 	idleTimeout, _ := time.ParseDuration(h.cfg.Session.IdleTimeout)
-	if idleTimeout > 0 && time.Since(sessionModel.LastUsed) > idleTimeout {
-		sessionDeletionErr := h.persister.GetSessionPersister().Delete(*sessionModel)
-		if sessionDeletionErr != nil {
-			return dto.ToHttpError(fmt.Errorf("failed to delete session: %w", sessionDeletionErr))
-		}
+		if idleTimeout > 0 && time.Since(sessionModel.LastUsed) > idleTimeout {
```
lfleischmann: Why indent here?

samadrehman:
closing this pr , not to assign any issue

PR #2647, opened 2026-06-03
feat: firebase scrypt compatibility

lfleischmann:
# Description

These changes provide Firebase Scryt compatibility. 

# Implementation

- Add command that converts firebase users exported via Firebase CLI to Hanko user import data
    - Passwords are stored in a specific format (inspired by Supabase) that allows recovering the Scrypt parameter/meta data on password verification
- Adjust password service password verification to include Firebase Scrypt
    -  The above mentioned formatted string is parsed, the Firebase specific Scrypt hash is created for a given input password and compared to the existing password hash. 

# How to test

1. Get the Firebase hash config of your Firebase project. Find it in the Firebase console under `Authentication > Users > three dots on the top right of the users table > Password hash parameters`.
2. Copy only the JSON object (don't include the `hash_config` string) and save it to a file.
3. Create one or more Firebase users, set a password.
4. Install the Firebase CLI, [log in](https://firebase.google.com/docs/cli#sign-in-test-cli), and then use the [auth:export](https://firebase.google.com/docs/cli/auth#auth-export) command for your project.
5. In the `backend` directory execute
    ```
    go run main.go user convert firebase --input="./firebase-users.json" --config="./hash-config.json"
    ```

    This will produce an output file (default `hanko-import.json`; can be changed via command option). Failed conversions are written into a "Dead Letter Queue" and saved in a separate output file (default `firebase-dlq.jsonl` where each line contains an error reason and the original firebase user; DLQ output file path can be changed via command option).
6. Run
    ```
    go run main.go user import --inputFile="./hanko-import.json"
    ```
7. Start the backend with a frontend example and try to log in with one of the user's credentials you chose when creating it in the Firebase console.

### DIFF
--- a/backend/cmd/user/firebase.go
+++ b/backend/cmd/user/firebase.go
@@ -401,33 +401,43 @@ func writer(
 
 func convertUser(u FirebaseUser, cfg FirebaseHashConfig) (ImportOrExportEntry, error) {
 
-	createdAt, _ := parseTime(u.CreatedAt)
-	updatedAt, _ := parseTime(u.UpdatedAt)
+	if u.Email == "" && u.PasswordHash != "" {
+		// If both are missing, there is no way to authenticate the user, unless some other credential/factor
+		// is manually added after the firebase -> hanko user conversion. This feels unlikely, so we count this
+		// as a failure.
+		return ImportOrExportEntry{}, fmt.Errorf("email and passwordHash missing for Firebase user with localId '%s'", u.LocalID)
+	}
 
 	fscryptString, err := buildFbscryptString(u, cfg)
 	if err != nil {
 		return ImportOrExportEntry{}, err
 	}
 
-	convertedEntry := ImportOrExportEntry{
-		Emails: []ImportOrExportEmail{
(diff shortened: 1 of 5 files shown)


backend/cmd/user/firebase.go
```diff
@@ -0,0 +1,545 @@
...
+		// If both are missing, there is no way to authenticate the user, unless some other credential/factor
+		// is manually added after the firebase -> hanko user conversion. This feels unlikely, so we count this
+		// as a failure.
+		return ImportOrExportEntry{}, fmt.Errorf("email and passwordHash missing for Firebase user with localId '%s'", u.LocalID)
+	}
+
+	fscryptString, err := buildFbscryptString(u, cfg)
+	if err != nil {
+		return ImportOrExportEntry{}, err
+	}
```
FreddyDevelop, line 447: As an optimization this could be done only when the `PasswordHash` ist not empty, e.g. it can move into the if clause on line 461.

FreddyDevelop (approved):

PR #2677, opened 2026-06-13
feat(i18n): add Korean (ko) translation

moduvoice:
# Description

Adds a **Korean (`ko`)** translation for the `hanko-elements` UI. Korean was the
only major East-Asian locale missing (the repo ships bn, de, en, fr, it, nl, pt-BR, zh).

# Implementation

- New `frontend/elements/src/i18n/ko.ts` — a full `Translation` (all keys present, matching `en.ts`), translated to natural Korean. Auth terminology is kept consistent (passkey → 패스키, passcode → 패스코드, security key → 보안 키, MFA → 다단계 인증(MFA), authenticator app → 인증 앱).
- Registered in `frontend/elements/src/i18n/all.ts`: imported `ko` and added it to the `all` map (alphabetically, between `it` and `nl`).
- All interpolation placeholders (`{emailAddress}`, `{minLength}`, `{maxLength}`, `{passwordRetryAfter}`, `{passcodeResendAfter}`, `{provider}`) are preserved 1:1 with the English source.

# Tests

Render any hanko element with `lang="ko"` (e.g. `<hanko-auth lang="ko">`) — all
headlines, texts, labels, and error messages display in Korean. Key parity with
`en.ts` was verified (181/181 keys, identical structure).


lfleischmann:
Hi @moduvoice,

this works well for the `all` dictionary but in order for the language to be accessible as a single scoped import (in case someone really insists on loading just this translation) it still needs an entry in the webpack configuration's `entry` map (see [here](https://github.com/teamhanko/hanko/blob/main/frontend/elements/webpack.config.cjs#L7-L78)).

moduvoice:
Thanks for catching that, @lfleischmann — added the `ko` entry to the webpack entry map so it can also be imported as a single scoped translation. Should be good to go now.

lfleischmann (approved):

lfleischmann:
@moduvoice Thanks, LGTM now, merging.

PR #2725, opened 2026-07-08
refactor(thirdparty): propagate context and fix duplicate defer in oauth providers

glatinone:
While evaluating hanko's OAuth integration sub-system, we identified two critical areas where production reliability and resource management under load can be significantly improved.

**1. Context Propagation Omission:**
- The `OAuthProvider` interface and its 7 implementations (GitHub, Apple, Discord, Microsoft, Facebook, LinkedIn, Custom) currently omit `context.Context`.
- This forces implementations to use `context.Background()` for outbound token exchanges and JWKS key fetches.
- Under heavy load or slow network responses, login handlers cannot cancel these blocked outbound requests upon client disconnect, potentially leading to goroutine accumulation and resource exhaustion.

**2. Double Defer and Silenced IO Read Error:**
- Inside the shared `makeRequest` helper in `provider.go`, `defer res.Body.Close()` is registered twice on the same HTTP body, and `io.ReadAll` errors are silently discarded via blank identifiers.
- This causes confusing behavior for static analyzers and risks propagating corrupt body bytes to the JSON decoder under partial network read failures.

**The Fix:**
- Threaded `context.Context` throughout the `OAuthProvider` interface, all 7 provider implementations (plus Google for interface completeness), and the shared `makeRequest` helper.
- Cleaned up the redundant defer registration in `makeRequest` and added robust error-checking for the body read execution path.
- Connected the incoming Echo request context directly to the OAuth orchestrator to enable clean cancellation propagation.

### DIFF
--- a/backend/handler/thirdparty.go
+++ b/backend/handler/thirdparty.go
@@ -116,12 +116,12 @@ func (h *ThirdPartyHandler) Callback(c echo.Context) error {
 		if state.CodeVerifier != "" && provider.ID() != "linkedin" {
 			opts = append(opts, oauth2.VerifierOption(state.CodeVerifier))
 		}
-		oAuthToken, terr := provider.GetOAuthToken(callback.AuthCode, opts...)
+		oAuthToken, terr := provider.GetOAuthToken(c.Request().Context(), callback.AuthCode, opts...)
 		if terr != nil {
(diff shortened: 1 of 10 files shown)


lfleischmann (approved):
LGTM, thanks.

PR #2739, opened 2026-07-13
fix: rate limit bypass

lfleischmann:
# Description

Rate limiting uses Echo's `c.RealIP()`. If the Echo IPExtractor is not explicitly configured, it trusts user-controlled headers such as `X-Forwarded-For` and `X-Real-IP`.

Said headers can be spoofed by a client, attackers can arbitrarily change their apparent IP address for each request, which resets the rate limit bucket.

This allows attackers to bypass authentication rate limits and perform unlimited brute-force attempts.

# Implementation

- Customizes echo IP Extractor: 
    - Can be configured using different IP extraction modes: `direct` (default) , `x_forwarded_for`, `x_real_ip` 
    - Trusted Proxy address ranges _must_ be configured if `x_forwarded_for` or `x_real_ip` modes are configured

# Tests

- Adds tests for the new IP Extractor customization (utility) itself
- Adds tests in `rate_limiter` package to test integration

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -25,6 +25,7 @@ easily integrated into any web app with as little as two lines of code.
   - [User import](#user-import)
   - [Webhooks](#webhooks)
   - [Session JWT templates](#session-jwt-templates)
+  - [Client IP address resolution](#client-ip-address-resolution)
 - [API specification](#api-specification)
 - [Configuration reference](#configuration-reference)
 - [License](#license)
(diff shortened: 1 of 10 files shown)


FreddyDevelop (approved):

FreddyDevelop:
We should document this in the Readme. This can have a bigger impact because when deployed behind a reverse-proxy and the ip-extractor is still set to `direct`, the rate limits can be exceeded quickly depending on the usage.

FreddyDevelop (approved):

PR #2759, opened 2026-07-20
fix: webauthn session data cred params

lfleischmann:
# Description

An earlier go-webauthn version added a mandatory algorithm-match check in `CreateCredential`, driven by `SessionData.CredParams`. Hanko round-trips WebAuthn session data through its own DB model instead of holding the
library struct in memory, and that mapping never carried `CredParams`, so every registration finish failed with "Credential public key algorithm not supported". Without this fix, we keep staying behind library updates.

# Implementation

- Persists `CredParams` via a new json column and round-trip it through `NewWebauthnSessionDataFrom`/`ToSessionData`. 
- Removes the orphaned `dto/intern/WebauthnSessionData.go` duplicate mapping, left over from the webauthn handler removal, which had the same gap and zero remaining callers.

# Tests

## Unit

- Adds small unit test for session data conversion 

## Manual

With and without these changes:
1. Start backend + frontend example
2. Create passkeys/security keys (registration + profile flow) and login with them (login flow).


FreddyDevelop:
Registration of a passkey works. But when using the passkey to login, an error occurs. In the logs I see `Backup Eligible flag inconsistency detected during login validation: this passkey cannot be used anymore`.
This only occurs when a client is used, that backups the keys, e.g. apple password manager or chrome profile.

The issue is, that since version [v0.11.0](https://github.com/go-webauthn/webauthn/releases/tag/v0.11.0) of the webauthn library, the backup flags are checked during login.

lfleischmann:
> Registration of a passkey works. But when using the passkey to login, an error occurs. In the logs I see `Backup Eligible flag inconsistency detected during login validation: this passkey cannot be used anymore`. This only occurs when a client is used, that backups the keys, e.g. apple password manager or chrome profile.
> 
> The issue is, that since version [v0.11.0](https://github.com/go-webauthn/webauthn/releases/tag/v0.11.0) of the webauthn library, the backup flags are checked during login.

Ugh great ... only tested with virtual authenticator in Chrome. Everything worked fine there.

PR #2826, opened 2026-08-11
feat: tenant scoped public user id

lfleischmann:
# Description

`users.id` (the DB primary key) is globally unique across all tenants, which breaks any scenario where two tenants legitimately want to use the same caller-supplied id (CLI import, admin API create-with-id). Pop's ORM has no support for composite primary keys, so `id` itself can't be relaxed without risking cross-tenant mutation/deletion via `Update()`/`Destroy()`.

This PR adds `users.public_id`, unique only per `(tenant_id, public_id)`, and makes it the **only** identifier ever exposed by any API surface - admin API, JWT `sub`, `/me`, webhooks, CLI import/export. The internal `id` becomes purely internal (FKs, WebAuthn handle, Pop mutations).

# Implementation

- Every existing row is backfilled (`public_id = id`) so pre-existing sessions/JWTs keep resolving after deploy; the column is `NOT NULL` at the DB level.
- Every **freshly created** user (self-registration, thirdparty sign-up, admin create without an explicit id, import without a `user_id`) gets an **independently generated** `public_id`, not a copy of `id` - this is a deliberate choice (see `models.NewUser`'s comment) so the public_id/id distinction is exercised by every new user, not just the admin/import minority that supplies an explicit id. Kept as one isolated, easily revertible commit.
- CLI import: re-importing a file with a `user_id` that already exists in the tenant now fails explicitly, instead of either silently upserting or crashing on a downstream unique-constraint collision.

Also fixes 4 correctness bugs and 4 cleanup items found via `/code-review` on this diff:
- metadata admin endpoints, email-limit check, and `hanko jwt create` were still resolving/comparing against the raw internal id instead of `public_id`.
- copy-paste error message, duplicated public_id fallback logic (consolidated into `User.GetPublicID()`), and `GetByPublicID` cut down from two DB round trips to one.

Not merged/deployed yet against MySQL/MariaDB or in a live WebAuthn/webhook end-to-end run - see manual test plan below.

# Test

Integration Tests included

## Manual

**1. Cross-tenant collision**
Enable multi-tenancy, seed two tenants. Create a user with the **same explicit `id`** in both via `POST /:tenant_id/users`. Both must succeed. `SELECT tenant_id, id, public_id FROM users WHERE public_id = '<uuid>'` should show two distinct internal `id`s, same `public_id`, different `tenant_id`.

**2. Backward compatibility for pre-existing data**
Run the migration against a copy of a real/staging DB. `SELECT count(*) FROM users WHERE public_id IS NULL OR public_id != id` must be zero. Take a JWT issued *before* this deploy and confirm it still validates against `/:tenant_id/sessions/validate` after deploying.

**3. The four correctness fixes, by hand**
- Metadata: create a user with an explicit `id`, `GET`/`PATCH /:tenant_id/users/:id/metadata`, confirm it resolves the right user.
- Email limit: set a low `emails.max_num_of_addresses`, add emails up to the limit for a user whose `public_id != id`, confirm the next one is rejected.
- `hanko jwt create`: run with a public id, decode the JWT (`sub` = public id), then check `SELECT user_id FROM sessions WHERE id = '<session id>'` - must hold the internal id.
- CLI import: import a file, re-import the same file into the same tenant - must fail with "already exists in this tenant" and leave zero partial state. Import into a **different** tenant - must succeed.

**4. JWT/session consistency across surfaces**
For one user, compare the `id` from `/me`, from `GET /:tenant_id/users/:id` (admin), and the `sub` claim in their session JWT - all three must match.

**5. WebAuthn end-to-end**
Real passkey registration + login in a browser .

**6. Other dialects**
All automated tests here ran against Postgres only. Run the migration up and down by hand against real MySQL and MariaDB instances, i.e. `migrate down` then `migrate up` again - confirm `public_id` drops and reapplies cleanly.

**7. Export/import round trip**
`export` a tenant, inspect the JSON (`user_id` must be the public id), then re-import that file into a **different** tenant - should succeed with fresh internal ids and preserved public ids.


backend/README.md
```diff
@@ -576,7 +576,10 @@ Unsafe metadata can be set and modified through the Public API and the Admin API
 
 
 ### User import
-You can import an existing user pool into Hanko using json in the following format:
+You can import an existing user pool into Hanko using json in the following format. `user_id`, if provided, only
+needs to be unique within the target tenant, not globally - if it already belongs to a user in that tenant, that
+user is reused rather than a duplicate being created, so the same file can be imported into multiple tenants (e.g.
+production and a staging/test tenant) without collision.
```
FreddyDevelop, line 582: The comment 
> that user is reused rather than a duplicate being created, so the same file can be imported into multiple tenants 

contradicts the description from `ImportExportEntry.UserID`, which states:

> If it already belongs to a user in this tenant, the import fails for this entry instead of reusing or overwriting that user.

FreddyDevelop (commented):
Just a quick look over the code, nothing tested yet.

backend/handler/metadata_admin_test.go
```diff
@@ -859,3 +859,35 @@ const metadataExceedingLimit = `{
     ]
   }
 }`
+
+// TestMetadataAdminHandler_UsesPublicIdNotInternalId is the regression test for the bug found in
+// review: GetMetadata/PatchMetadata were never migrated to public_id lookups, unlike every
+// sibling handler mounted under the same /users/:id route group.
```
FreddyDevelop: Mentioning a review, but it is not a helpful information.

backend/handler/email_admin_test.go
```diff
@@ -245,6 +245,50 @@ func (s *emailAdminSuite) TestEmailAdminHandler_Create() {
 	}
 }
 
+// TestEmailAdminHandler_Create_EmailLimitUsesResolvedInternalId is the regression test for the
+// email-limit bypass found in review: the limit check counted emails by the raw path variable
```
FreddyDevelop:
```suggestion
// TestEmailAdminHandler_Create_EmailLimitUsesResolvedInternalId is the regression test for a
// email-limit bypass: the limit check counted emails by the raw path variable
```

backend/handler/email_admin_test.go
```diff
@@ -525,3 +569,41 @@ func (s *emailAdminSuite) TestEmailAdminHandler_SetPrimaryEmail() {
 		})
 	}
 }
+
+// TestEmailAdminHandler_List_UsesPublicIdNotInternalId closes the one gap flagged in review:
+// List is the only email_admin.go handler that didn't already pre-fetch the user before this
+// change, so it's the one most worth a dedicated regression test proving :user_id resolves via
+// public_id and not the real internal id.
```
FreddyDevelop: Also mentioning a review.

backend/handler/email_admin_test.go
```diff
@@ -245,6 +245,50 @@ func (s *emailAdminSuite) TestEmailAdminHandler_Create() {
 	}
 }
 
+// TestEmailAdminHandler_Create_EmailLimitUsesResolvedInternalId is the regression test for the
+// email-limit bypass found in review: the limit check counted emails by the raw path variable
+// (public_id) instead of the resolved internal user id, so it always read 0 and never tripped for
+// any user whose public_id differs from their internal id. A limit of 0 (as in the existing
+// "should reject" test case above) can't distinguish buggy from fixed behavior, since 0 >= 0 is
+// true either way - this needs a non-zero limit and a user who already has an email.
```
FreddyDevelop: This also mentions a review.

FreddyDevelop (approved):
Looks good. Only some comments for the tests should be changed, because they mention some review. But it is no blocker.

FreddyDevelop (approved):

PR #2836, opened 2026-08-12
500 flow_discontinuity_error on login when account doesn't exist and ShowAccountExistenceHints is false

tsushanth:
add nil userModel guard after ShowAccountExistenceHints block so unknown email with privacy hints disabled returns clean FlowDiscontinuity instead of 500

Found via automated repo scanning, fix written and reviewed before opening.


FreddyDevelop:
That is intended behaviour. When ShowAccountExistenceHints is disabled, the application should not behave differently, depending if the user exists or not. For privacy reason some applications do not want to disclose if an account exist.

PR #2882, opened 2026-09-05
fix(backend): allow long user agents

### LINKED ISSUE #2877: Login and registration return 500 when the User-Agent exceeds 255 characters
### Checklist

### Describe the bug

Login and registration return a 500 when the client sends a User-Agent longer than 255 characters, with `session.acquire_user_agent` on (the default). The password or passcode is verified first, so the visitor answers correctly and then gets "Something went wrong" with no session.

The session insert fails with:

```
the action failed to handle the request: failed to execute before state hook (state: success): failed to store session: failed to store session: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
```

I hit it in production from the Instagram in-app browser on Android, which sends 278 characters:

```
Mozilla/5.0 (Linux; Android 16; SM-S911B Build/BP4A.251205.006; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/152.0.7977.69 Mobile Safari/537.36 Instagram 445.0.0.45.83 Android (36/16; 480dpi; 1080x2340; samsung; SM-S911B; dm1q; qcom; nl_NL; 1055488490; IABMV/1)
```

I have not debugged the backend, so the cause below is a guess from reading the source. `sessions.user_agent` is declared as an unsized fizz `string` in [20241002113000_create_sessions.up.fizz](https://github.com/teamhanko/hanko/blob/main/backend/persistence/migrations/20241002113000_create_sessions.up.fizz#L4), which fizz maps to `VARCHAR (255)` on Postgres ([translators/postgres.go](https://github.com/gobuffalo/fizz/blob/v1.14.4/translators/postgres.go#L211-L218)), and the header is stored verbatim in [hook_issue_session.go](https://github.com/teamhanko/hanko/blob/main/backend/flow_api/flow/shared/hook_issue_session.go#L109-L113). Whether the fix is truncating before the insert, widening the column, or something else is your call.

### Reproducing the bug

1. Run the backend against Postgres with `session.acquire_user_agent: true`.
2. Complete a login or a registration from a client sending a User-Agent over 255 characters against the flow API, or open the app in an Instagram or Facebook in-app browser.
3. The `before` hook of the `success` state fails and the request returns 500 `technical_error`. No session row is written and no token is issued.

I reproduced this locally with the 278-character string above on both `password_login` and registration. A 111-character User-Agent on the same build succeeds, and so does any login with `acquire_user_agent: false`.

### Logs

```Shell
POST /login?action=password_login@...  status 500  error technical_error
the action failed to handle the request: failed to execute before state hook (state: success): failed to store session: failed to store session: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
```

### Configuration

```yml
session:
  acquire_user_agent: true
```

### Hanko Version

v2.7.0

### OS Hanko Backend

Linux

### OS Version Hanko Backend

Docker Compose, backend on Linux, Postgres. Client: Instagram in-app browser on Android 16.

### OS

Other

### OS Version

_No response_

### Browser Version

_No response_

### Environment

None

### Additional Context

I work around it by trimming the User-Agent to 255 characters in the reverse proxy in front of Hanko. Setting `session.acquire_user_agent: false` also avoids it, at the cost of the user agent on the profile session list.

`audit_logs.meta_user_agent` is declared the same way in [20220818111000_create_audit_logs.up.fizz](https://github.com/teamhanko/hanko/blob/main/backend/persistence/migrations/20220818111000_create_audit_logs.up.fizz), so the same overflow looks reachable there.

ousamabenyounes:
# Description
Fix #2877.

Login and registration can persist the request User-Agent in sessions and audit logs. Those columns were backed by Fizz `string` columns, which map to a 255-character limit and reject longer in-app browser headers.

# Implementation
- Change `sessions.user_agent` and `audit_logs.meta_user_agent` to `text` columns.
- Keep the existing nullable setting for `sessions.user_agent`.
- Make the down migration rollback-safe by truncating values longer than the old string column limit before restoring the previous column type.

## Test verification (RED -> GREEN)

RED on upstream `main` with only the regression test added:

```text
failed to store auditlog: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
failed to store session: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
FAIL github.com/teamhanko/hanko/backend/v3/persistence
```

GREEN after the migration change:

```text
--- PASS: TestUserAgentSuite (4.47s)
    --- PASS: TestUserAgentSuite/TestAuditLogPersister_CreateStoresLongUserAgent (0.79s)
    --- PASS: TestUserAgentSuite/TestSessionPersister_CreateStoresLongUserAgent (0.77s)
PASS
ok github.com/teamhanko/hanko/backend/v3/persistence 4.493s
```

Regression check with the migration removed again:

```text
failed to store auditlog: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
failed to store session: named insert: ERROR: value too long for type character varying(255) (SQLSTATE 22001)
FAIL github.com/teamhanko/hanko/backend/v3/persistence
```

Full local suite proof:

```text
cd backend && go generate ./... && go build -v ./... && go test -v ./...
PASS
ok github.com/teamhanko/hanko/backend/v3/persistence 6.587s
```

# Todos
None.

# Additional context
Validation ran in Docker with Go 1.26 and the repo's dockertest-backed PostgreSQL setup.

### DIFF
--- a/backend/persistence/user_agent_test.go
+++ b/backend/persistence/user_agent_test.go
@@ -1,0 +1,87 @@
+package persistence_test
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gobuffalo/nulls"
+	"github.com/gofrs/uuid"
+	"github.com/stretchr/testify/suite"
+	"github.com/teamhanko/hanko/backend/v3/config"
+	"github.com/teamhanko/hanko/backend/v3/persistence/models"
+	"github.com/teamhanko/hanko/backend/v3/test"
+)
+
+const (
+	longUserAgentCharacter = "a"
+	longUserAgentLength    = 278
+	testSourceIP           = "127.0.0.1"
+)
+
+func TestUserAgentSuite(t *testing.T) {
+	t.Parallel()
+	suite.Run(t, new(userAgentSuite))
+}
+
+type userAgentSuite struct {
+	test.Suite
+}
+
+func (s *userAgentSuite) TestSessionPersister_CreateStoresLongUserAgent() {
+	if testing.Short() {
+		s.T().Skip("skipping test in short mode.")
+	}
+
+	tenantID := uuid.FromStringOrNil(config.DefaultTenantID)
+	user := models.NewUser(tenantID)
+	err := s.Storage.GetUserPersister().Create(user)
+	s.Require().NoError(err)
+
+	userAgent := strings.Repeat(longUserAgentCharacter, longUserAgentLength)
+	now := time.Now().UTC()
+	session := models.Session{
+		ID:        uuid.Must(uuid.NewV4()),
+		UserID:    user.ID,
+		TenantID:  tenantID,
+		UserAgent: nulls.NewString(userAgent),
+		CreatedAt: now,
+		UpdatedAt: now,
+		LastUsed:  now,
+	}
+
+	err = s.Storage.GetSessionPersister().Create(session)
+	s.Require().NoError(err)
+
+	stored, err := s.Storage.GetSessionPersister().Get(session.ID, tenantID)
+	s.Require().NoError(err)
+	s.Require().NotNil(stored)
+	s.Equal(userAgent, stored.UserAgent.String)
+}
+
+func (s *userAgentSuite) TestAuditLogPersister_CreateStoresLongUserAgent() {
+	if testing.Short() {
+		s.T().Skip("skipping test in short mode.")
+	}
+
+	userAgent := strings.Repeat(longUserAgentCharacter, longUserAgentLength)
+	now := time.Now().UTC()
+	auditLog := models.AuditLog{
+		ID:                uuid.Must(uuid.NewV4()),
+		Type:              models.AuditLogLoginSuccess,
+		MetaHttpRequestId: uuid.Must(uuid.NewV4()).String(),
+		MetaSourceIp:      testSourceIP,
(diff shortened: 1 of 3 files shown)


backend/persistence/user_agent_test.go
lfleischmann, line 1: I'd rather not have feature based test files and follow the pattern in the `handler` directory instead, i.e. test files per persister. Can you split this up accordingly?

lfleischmann (changes requested):

ousamabenyounes:
Thanks for the review, @lfleischmann. Addressed in `ca21ecb`:

- Dropped the feature-based `backend/persistence/user_agent_test.go`.
- Split its cases into per-persister files following the `handler` pattern:
  - `session_persister_test.go` → `TestSessionPersister_CreateStoresLongUserAgent`
  - `audit_log_persister_test.go` → `TestAuditLogPersister_CreateStoresLongUserAgent`

Behavior is unchanged; only the test layout moved. CI is green on `ca21ecb5781ab6a73e5391a8498047adbd0f4207`.

lfleischmann (approved):

lfleischmann:
LGTM now. Thanks.

PR #2883, opened 2026-09-05
fix(frontend-sdk): bound passkey creation with a ceremony deadline

### LINKED ISSUE #2867: hanko-auth: user permanently stuck on passkey onboarding when navigator.credentials.create() never settles (Skip button unusable)
### Checklist

### Describe the bug

Environment: Hanko Cloud, @teamhanko/hanko-elements 2.7.0. Observed with Safari (macOS and iOS); the underlying trigger is browser-specific but the recovery flaw is in the component.

Description

On the onboarding_create_passkey step, pressing "Create a passkey" runs the webauthn_generate_creation_options action and the flow moves to webauthn_credential_verification. The component keeps the previous page rendered with all controls disabled while the state's auto-step awaits navigator.credentials.create() (via the WebauthnManager).

If the promise rejects, recovery works: the auto-step runs back and shows an error. But if the promise never settles, the component stays disabled forever — the Skip button is visible but dead, and the user cannot proceed, skip, or leave the step.

A never-settling create() is not hypothetical: Safari's platform authenticator can get stuck (for example right after a passkey was used in another app), and only fully quitting and restarting Safari clears it. We lost a real signup to this before disabling passkey onboarding from our Cloud console as a workaround.

### Reproducing the bug

Steps to reproduce (deterministic simulation)

Open a registration flow and proceed to the "Create a passkey" onboarding step.
In the console, stub the ceremony so it hangs: navigator.credentials.create = () => new Promise(() => {})
Click "Create a passkey".
Observe: the page stays on the passkey step with every control disabled; Skip never becomes clickable again. No timeout ever fires.

Expected behavior

Either Skip/Back remain usable during the ceremony, or the ceremony has a deadline (e.g. AbortController + timeout in the auto-step) after which the flow returns to onboarding_create_passkey with an error and the controls re-enable.

Workaround we applied: disabled passkey acquisition at registration/login in the Cloud console ("Prompt during registration/login: Never"), keeping passkey creation available from the profile.

### Logs

```Shell

```

### Configuration

```yml

```

### Hanko Version

hanko-elements 2.7.0 (backend: Hanko Cloud SaaS)

### OS Hanko Backend

None

### OS Version Hanko Backend

Hanko Cloud (SaaS)

### OS

Other

### OS Version

iOS (recent; exact version not recorded). Also observed on macOS Safari.

### Browser Version

Safari (iOS and macOS). The hang requires Safari's platform authenticator; the missing recovery/timeout is browser-independent.

### Environment

None

### Additional Context

Reported first via Hanko Cloud support (they asked us to file this issue). We verified the mechanism by reading the bundled elements 2.7.0 code. A restart of Safari clears the stuck authenticator. A separate related issue about the login flow resetting mid-passcode is being filed next.

ousamabenyounes:
# Description

Fixes #2867

On `onboarding_create_passkey` (and in the profile's passkey section) the component keeps the previous page rendered with every control disabled while the auto-step awaits `navigator.credentials.create()`. A ceremony that *rejects* recovers fine — the auto-step runs `back` and the controls come back. A ceremony that never settles does not: nothing ever ends it, so the step stays disabled and the user can neither create a passkey nor use Skip.

That is reachable in practice. Safari's platform authenticator can get stuck (WebKit bug 273712) and Safari does not apply the WebAuthn `timeout` itself, so the promise stays pending until the browser is fully restarted.

# Implementation

`WebauthnManager.createWebauthnCredential()` now races the ceremony against a deadline:

- the deadline is the `timeout` the API already sends in the creation options, falling back to `DEFAULT_CREATION_TIMEOUT_MS` (600000 ms, the same value as the backend's `webauthn.timeouts.registration` default). No new configuration is introduced — this just enforces client-side the bound that setting already describes. A zero timeout is treated as unset rather than as "abort now", so a `webauthn.timeouts.registration: 0` deployment does not end up unable to register a passkey at all;
- when it elapses, the request's own `AbortController` is aborted and the promise rejects with the existing `RequestTimeoutError`;
- the timer is cleared in a `finally`, so a ceremony that completes normally is never aborted afterwards;
- the controller is captured per request, so a deadline can only abort its own ceremony and never a later one that superseded it.

`handleCredentialCreation()` in `auto-steps.ts` maps that rejection to a `technical_error` flow error. Both credential-creation entry points — `onboarding_verify_passkey_attestation` (registration) and `webauthn_credential_verification` (profile) — go through that helper, so the step is re-entered with a visible message instead of silently.

`getWebauthnCredential()` and `getConditionalWebauthnCredential()` are deliberately untouched: conditional mediation is meant to stay armed for the lifetime of the page, and a deadline there would break passkey autofill.

# Tests

```
cd frontend/frontend-sdk
npx jest tests/lib/flow-api/WebauthnManager.spec.ts tests/lib/flow-api/Webauthn.spec.ts
```

Six new tests: four cover the added branches (deadline from the creation options, deadline from the fallback, a superseded ceremony, and the flow error surfaced by the auto-step) and fail without the change; two guard against getting the change wrong — a ceremony finishing before the deadline must not be aborted afterwards, and a zero timeout must not end the ceremony immediately.

## Test verification (RED → GREEN)

**RED** — on `main`, with only the new tests applied and no production change:

```
  ● WebauthnManager credential creation deadline › rejects with a timeout error when the ceremony never settles

    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 1

      Object {
    -   "error": Any<RequestTimeoutError>,
    -   "status": "rejected",
    +   "status": "pending",
      }

  ● autoSteps webauthn credential creation error handling › surfaces an error when the ceremony times out

    TypeError: Cannot read properties of undefined (reading 'code')

      75 |     expect(state.actions.back.run).toHaveBeenCalled();
    > 76 |     expect(result.error.code).toBe("technical_error");

Test Suites: 2 failed, 2 total
Tests:       4 failed, 6 passed, 10 total
```

**GREEN** — with the fix applied:

```
PASS tests/lib/flow-api/WebauthnManager.spec.ts
PASS tests/lib/flow-api/Webauthn.spec.ts

Test Suites: 2 passed, 2 total
Tests:       10 passed, 10 total
```

Full suite (`cd frontend && npm ci && npm run build && npm test`) goes from 97 to 103 passing tests with no other change, and the frontend build is unaffected.

Also checked in a real browser (headless Chromium via Playwright), stubbing the ceremony exactly as the issue describes — `navigator.credentials.create = () => new Promise(() => {})` — with a 3000 ms `timeout` in the creation options and an 8000 ms observation window:

```
main:      outcome: {"status":"pending"}
this PR:   outcome: {"status":"rejected","code":"requestTimeout","elapsedMs":3004}
```

# Additional context

`technical_error` is an existing `flowErrors` key present in every `hanko-elements` locale, so the timeout surfaces a translated message ("A technical error has occurred. Please try again later.") without any i18n change.

### DIFF
--- a/frontend/frontend-sdk/src/lib/flow-api/WebauthnManager.ts
+++ b/frontend/frontend-sdk/src/lib/flow-api/WebauthnManager.ts
@@ -6,6 +6,12 @@ import {
   create,
   get,
 } from "@github/webauthn-json";
+import { RequestTimeoutError } from "../Errors";
+
+// Applied when the creation options carry no usable `timeout`. Mirrors the
+// Hanko API's own `webauthn.timeouts.registration` default, so the deadline
+// enforced here matches the one the server would have asked for.
+const DEFAULT_CREATION_TIMEOUT_MS = 600000;
 
 /**
  * Manages WebAuthn credential operations as a singleton, ensuring only one active request at a time.
@@ -76,17 +82,46 @@ class WebauthnManager {
   /**
    * Creates a new WebAuthn credential using the provided options.
    * Aborts any previous request before starting a new one.
+   *
+   * The ceremony is bounded by the `timeout` given in the creation options
+   * (falling back to {@link DEFAULT_CREATION_TIMEOUT_MS}). Some authenticators
+   * leave `navigator.credentials.create()` pending indefinitely and do not
+   * honor the WebAuthn `timeout` themselves, which would keep the caller
+   * waiting forever; enforcing the deadline here turns that into a regular
+   * rejection the caller can recover from.
    * @param {CredentialCreationOptionsJSON} options - The options for credential creation
    * @returns {Promise<PublicKeyCredentialWithAttestationJSON>} A promise resolving to the created credential
    * @throws {DOMException} If the WebAuthn request fails (e.g., aborted, not allowed)
+   * @throws {RequestTimeoutError} If the ceremony does not complete before the deadline
    */
   public async createWebauthnCredential(
     options: CredentialCreationOptionsJSON,
   ): Promise<PublicKeyCredentialWithAttestationJSON> {
-    return await create({
-      ...options,
-      signal: this.createAbortSignal(),
-    });
+    const signal = this.createAbortSignal();
+    // createAbortSignal() has just installed a fresh controller; hold on to it
+    // so the deadline below aborts this ceremony rather than a later one.
+    const controller = this.abortController;
+    // A missing or zero `timeout` is not a request to end the ceremony at once
+    // - the WebAuthn timeout is a hint that clients clamp anyway - so fall back
+    // to the default instead of aborting before the user can even respond.
+    const requestedTimeout = options.publicKey?.timeout;
+    const timeout =
+      requestedTimeout > 0 ? requestedTimeout : DEFAULT_CREATION_TIMEOUT_MS;
+    let deadline: ReturnType<typeof setTimeout>;
+
+    try {
+      return await Promise.race([
+        create({ ...options, signal }),
+        new Promise<never>((_resolve, reject) => {
+          deadline = setTimeout(() => {
+            controller.abort();
+            reject(new RequestTimeoutError());
+          }, timeout);
+        }),
+      ]);
+    } finally {
+      clearTimeout(deadline);
+    }
   }
 }
 
--- a/frontend/frontend-sdk/src/lib/flow-api/auto-steps.ts
+++ b/frontend/frontend-sdk/src/lib/flow-api/auto-steps.ts
@@ -3,6 +3,11 @@ import { WebauthnSupport } from "../WebauthnSupport";
 import WebauthnManager from "./WebauthnManager";
 import { CredentialCreationOptionsJSON } from "@github/webauthn-json";
 import { clearStoredCodeVerifier, getStoredCodeVerifier } from "../Pkce";
+import { RequestTimeoutError } from "../Errors";
+
+// Generic flow error code, translated by hanko-elements as "A technical error
+// has occurred. Please try again later."
+const TECHNICAL_ERROR_CODE = "technical_error";
 
 // Helper function to handle WebAuthn credential creation and error handling
 // eslint-disable-next-line require-jsdoc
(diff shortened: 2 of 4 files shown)


frontend/frontend-sdk/src/lib/flow-api/auto-steps.ts
```diff
@@ -28,6 +33,10 @@ async function handleCredentialCreation(
     // the login_passkey step.
     if (error instanceof DOMException && error.name === "InvalidStateError") {
       nextState.error = { code: errorCode, message: errorMessage };
+    } else if (error instanceof RequestTimeoutError) {
+      // The authenticator never answered. Returning without an error would drop
+      // the user back on the passkey step with no explanation for the restart.
+      nextState.error = { code: TECHNICAL_ERROR_CODE, message: error.message };
```
lfleischmann, line 39: Rather than surfacing this as a generic technical error, I would prefer staying true to the error at hand:  add a dedicated `flowErrors` key (e.g. `request_timeout`) across all locale files in `frontend/elements/src/i18n` and use that instead.

frontend/frontend-sdk/src/lib/flow-api/WebauthnManager.ts
```diff
@@ -6,6 +6,12 @@ import {
   create,
   get,
 } from "@github/webauthn-json";
+import { RequestTimeoutError } from "../Errors";
+
+// Applied when the creation options carry no usable `timeout`. Mirrors the
+// Hanko API's own `webauthn.timeouts.registration` default, so the deadline
+// enforced here matches the one the server would have asked for.
+const DEFAULT_CREATION_TIMEOUT_MS = 600000;
```
lfleischmann, line 15: `DEFAULT_CREATION_TIMEOUT_MS` and the `requestedTimeout > 0 ? ... : DEFAULT_CREATION_TIMEOUT_MS` check are rather redundant against any real Hanko backend: it always returns a non-zero `publicKey.timeout` — either Hanko's own default (600000) or, if you set `registration: 0` in the config, `go-webauthn`'s internal 300000 fallback (substituted before the JSON is even built, so the field is never absent or zero on the wire). So this fallback path is practically unreachable for the actual product.

lfleischmann (changes requested):

ousamabenyounes:
Thanks for the detailed review, @lfleischmann. Addressed in `bb6438a`:

1. Added a dedicated `request_timeout` key to `flowErrors` across all `frontend/elements/src/i18n` locales (and its type in `translations.ts`), and surface it for `RequestTimeoutError` instead of the generic `technical_error`.
2. Dropped the redundant `DEFAULT_CREATION_TIMEOUT_MS` fallback: since a real Hanko backend always sends a non-zero `publicKey.timeout`, the deadline now uses it directly.
3. Updated the affected tests accordingly (expect `request_timeout`; removed the obsolete zero-timeout fallback case).

CI is green on `bb6438a56b7a862283ba6f3727ccdb2ad77c0248`.

PR #2884, opened 2026-09-08
feat: tenant-defined custom claims mapped from SAML/OIDC connections

lfleischmann:
# Description

Adds tenant-defined "custom claims": SAML and custom OIDC/OAuth2 connections can map their own attributes/claims onto a tenant-declared, typed set of claim names.

# Implementation

- `custom_claims.definitions`: name -> `{type, description}`, scalar/list-of-scalar types only, capped at 50.
- Mapping: SAML `attribute_map.custom`, OIDC/custom providers' `custom_claim_mapping` (gjson path into raw claims). Validated against declared definitions on save, and on definition removal.
- Resolution at login: last-one-wins, source-scoped clear (only the owning connection can clear a claim), writes/webhook gated on actual value change.
- New `user_custom_claims` table, one row per user/tenant.
- Exposed via `.User.CustomClaims "name"` in JWT templates - preserves the real type (number, boolean, array) when the claim's template value is that reference alone, e.g. `'{{ .User.CustomClaims "age" }}'`, not glued to other text - and a read-only `GET /users/:id/custom_claims` Admin API.
- New `user.update.custom_claims` webhook event.

Also included: new `hanko jwk create --store` (persists a JWK for a tenant instead of just printing one), with `--if-missing` to skip instead of creating an extra key if the tenant already has one; removal of a dead pre-existing leftover-claims mechanism; and kind-cluster deploy tooling for local testing.

# Tests

**Setup**: `docker compose -f quickstart.yaml -f quickstart.keycloak.yaml up -d` in `deploy/docker-compose` - brings up Hanko plus a Keycloak instance with two SAML realms (`uni-a`, `uni-b`) and one OIDC client (`acme-corp`), already configured with custom claim mappings. Also starts up a small "webhook receiver" that just logs webhook payloads on incoming requests, so the new webhook event can be observed.

**Major test cases**:
- Declare a claim, map it on a SAML connection's `attribute_map.custom`, log in via that IdP, confirm the resolved value in the session JWT and via `GET /users/:id/custom_claims`.
- Same, but mapped on an OIDC/custom provider's `custom_claim_mapping`.
- Log in a second time via a *different* IdP asserting the same claim - confirm ownership transfers (source updates) but no `user.update.custom_claims` webhook fires, since the value itself didn't change.
- Change a custom claim value at the provider, observe that `user.update.custom_claims` webhook fires and payload data contains updated value
- Have the owning connection stop asserting a claim - confirm it gets cleared. Have an unrelated connection go silent on a claim it never owned - confirm it's untouched.
- Try mapping a connection to an undeclared claim name - confirm it's rejected with a real error message, not a generic one.
- Try deleting a claim definition while a SAML connection still maps it - confirm it's rejected.


backend/handler/user_custom_claims_admin.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get tenant from context: %w", err)
+	}
+
+	userID, err := uuid.FromString(c.Param("id"))
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "invalid user id")
+	}
+
+	userExists, err := h.persister.GetConnection().Where("id = ?", userID).Exists(&models.User{ID: userID})
```
FreddyDevelop: The user-existence check is not scoped to the current tenant.

backend/persistence/models/user_custom_claims.go
```diff
@@ -0,0 +1,52 @@
...
+	ID        uuid.UUID    `db:"id"`
+	UserID    uuid.UUID    `db:"user_id"`
+	TenantID  uuid.UUID    `db:"tenant_id"`
+	Claims    nulls.String `db:"claims"`
+	CreatedAt time.Time    `db:"created_at"`
+	UpdatedAt time.Time    `db:"updated_at"`
+}
+
+func (c *UserCustomClaims) Validate(_ *pop.Connection) (*validate.Errors, error) {
+	claimsMax := 3000
```
FreddyDevelop: Increase the max claims size to `10000` so we would likely never run into the limit, because when it is reached the login for a user will fail.

backend/dto/admin/custom_claims.go
```diff
@@ -0,0 +1,39 @@
...
+// governs which connection may clear it.
+type CustomClaims map[string]any
+
+// NewCustomClaims unwraps the stored {claimName: {value, source}} envelope into a flat
+// {claimName: value} view. Returns nil (leading to 204 No Content, mirroring metadata) if
+// the user has no custom claims at all.
+func NewCustomClaims(model *models.UserCustomClaims) (CustomClaims, error) {
+	if !model.Claims.Valid || model.Claims.String == "" {
+		return nil, nil
+	}
```
FreddyDevelop:
```suggestion
	if customClaims == nil || !model.Claims.Valid || model.Claims.String == "" {
		return nil, nil
	}
```

backend/persistence/user_custom_claims_persister.go
```diff
@@ -0,0 +1,97 @@
...
+func (e *CustomClaimsLimitExceededError) Unwrap() error {
+	return e.ValidationErrors
+}
+
+// IsCustomClaimsLimitExceededError checks if the error is a CustomClaimsLimitExceededError
+func IsCustomClaimsLimitExceededError(err error) bool {
+	var customClaimsLimitExceededError *CustomClaimsLimitExceededError
+	ok := errors.As(err, &customClaimsLimitExceededError)
+	return ok
+}
```
FreddyDevelop: This function is never used.

backend/thirdparty/custom_claims.go
```diff
@@ -0,0 +1,268 @@
...
+// what tells apart "mapped but the IdP sent nothing this login" (a clear candidate) from both
+// "this connection doesn't map this claim at all" and "mapped, but this round's value didn't
+// coerce" (both invisible to this call entirely, since neither is evidence the claim no longer
+// applies). A managed claim missing from resolvedValues is therefore always the genuine-absence
+// case.
+//
+// wrote reports whether stored was mutated at all, including a source-only handoff (another
+// connection re-asserting a value this claim already had) - that must still persist so
+// ownership stays correct for future clears. valueChanged reports whether a claim's value
+// itself appeared, disappeared, or changed - false for a source-only handoff.
```
FreddyDevelop:
```suggestion
// `wrote` reports whether `stored` was mutated at all, including a source-only handoff (another
// connection re-asserting a value this claim already had) - that must still persist so
// ownership stays correct for future clears. `valueChanged` reports whether a claim's value
// itself appeared, disappeared, or changed - false for a source-only handoff.
```

backend/thirdparty/custom_claims.go
```diff
@@ -0,0 +1,268 @@
...
+
+// applyCustomClaims resolves userData.CustomClaimSource and merges it into the user's
+// persisted custom claims. Last-one-wins, any source; a claim found no value for is cleared
+// only if this same connection was its stored source.
+//
+// record is non-nil whenever anything was written - including a source-only handoff (another
+// connection re-asserting a value this claim already had), which still needs persisting so
+// ownership stays correct for future clears. valueChanged is true only when a claim's value
+// itself appeared, disappeared, or changed - gate a user.update webhook on this, not on
+// record != nil, so re-asserting an unchanged value never fires one.
```
FreddyDevelop:
```suggestion
// `record` is non-nil whenever anything was written - including a source-only handoff (another
// connection re-asserting a value this claim already had), which still needs persisting so
// ownership stays correct for future clears. `valueChanged` is true only when a claim's value
// itself appeared, disappeared, or changed - gate a user.update webhook on this, not on
// record != nil, so re-asserting an unchanged value never fires one.
```

deploy/docker-compose/config-keycloak.yaml
```diff
@@ -0,0 +1,187 @@
...
+  cookie:
+    secure: false # is needed for safari, because safari does not store secure cookies on localhost
+  jwt_template:
+    claims:
+      matriculation_number: '{{ .User.CustomClaims.Get "matriculation_number" }}'
+      affiliation: '{{ .User.CustomClaims.Get "affiliation" }}'
+      account_status: '{{ .User.CustomClaims.Get "account_status" }}'
+      email_verified: '{{ .User.CustomClaims.Get "email_verified" }}'
+      employee_id: '{{ .User.CustomClaims.Get "employee_id" }}'
+      home_organization: '{{ .User.CustomClaims.Get "home_organization" }}'
```
FreddyDevelop:
```suggestion
      matriculation_number: '{{ .User.CustomClaims "matriculation_number" }}'
      affiliation: '{{ .User.CustomClaims "affiliation" }}'
      account_status: '{{ .User.CustomClaims "account_status" }}'
      email_verified: '{{ .User.CustomClaims "email_verified" }}'
      employee_id: '{{ .User.CustomClaims "employee_id" }}'
      home_organization: '{{ .User.CustomClaims "home_organization" }}'
```

deploy/docker-compose/quickstart.keycloak.yaml
```diff
@@ -0,0 +1,74 @@
...
+  hanko-migrate:
+    volumes:
+      - type: bind
+        source: ./config-keycloak.yaml
+        target: /etc/config/config.yaml
+  hanko:
+    volumes:
+      - type: bind
+        source: ./config-keycloak.yaml
+        target: /etc/config/config.yaml
```
FreddyDevelop, line 35:
```suggestion
  hanko:
    depends_on:
      keycloak: 
        condition: service_healthy
      hanko-migrate:
        condition: service_completed_successfully
    volumes:
      - type: bind
        source: ./config-keycloak.yaml
        target: /etc/config/config.yaml
```

deploy/docker-compose/quickstart.keycloak.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - intranet
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.0
+    depends_on:
+      keycloak-db-init:
+        condition: service_completed_successfully
+    # --import-realm loads every *-realm.json file found in the mounted import
+    # dir on startup, so a realm exported once and committed there is available
+    # to every dev without redoing the manual setup in the admin console.
+    command: start-dev --import-realm
```
FreddyDevelop:
```suggestion
    command: start-dev --import-realm --health-enabled=true
```

deploy/docker-compose/quickstart.keycloak.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # hostname into browser-facing SAML redirects.
+      - KC_HOSTNAME=http://localhost:9000
+    ports:
+      - '9000:8080' # admin console + SAML/OIDC endpoints
+    volumes:
+      - type: bind
+        source: ./keycloak-import
+        target: /opt/keycloak/data/import
+    networks:
+      - intranet
```
FreddyDevelop, line 84:
```suggestion
    volumes:
      - type: bind
        source: ./keycloak-import
        target: /opt/keycloak/data/import
    healthcheck:
        test: [ 'CMD-SHELL', '[ -f /tmp/HealthCheck.java ] || echo "public class HealthCheck { public static void main(String[] args) throws java.lang.Throwable { System.exit(java.net.HttpURLConnection.HTTP_OK == ((java.net.HttpURLConnection)new java.net.URL(args[0]).openConnection()).getResponseCode() ? 0 : 1); } }" > /tmp/HealthCheck.java && java /tmp/HealthCheck.java http://localhost:9000/health/ready' ]
        interval: 5s
        timeout: 5s
        retries: 30
    networks:
      - intranet
```

backend/persistence/user_custom_claims_persister.go
lfleischmann:
```suggestion
```

backend/persistence/models/user_custom_claims.go
lfleischmann:
```suggestion
	claimsMax := 10000
```

FreddyDevelop (approved):

FreddyDevelop (approved):

PR #2893, opened 2026-09-10
feat: organizations and roles

lfleischmann:
# Description

Adds organizations and roles: users can be grouped into organizations within a tenant, tenant-wide roles can be bound to a user within a specific organization, and a public endpoint lets a relying party check whether the current session holds one of a given set of roles in an organization. Existing endpoints (`POST /users`, `POST /sessions/validate`, `GET /users`) are extended to create and surface this data.

# Implementation

**New endpoints:**
- Organizations: `POST/GET/PATCH/DELETE /organizations[/{id}]`
- Organization membership: `POST/DELETE /organizations/{org_id}/users/{user_id}`, `GET /organizations/{org_id}/users`
- Roles: `POST/GET/PATCH/DELETE /roles[/{id}]`
- Role bindings: `POST/GET /organizations/{org_id}/users/{user_id}/roles`, `DELETE .../roles/{role_id_or_slug}`
- Public: `POST /organizations/roles/check` — role check for the current session

**Integration with existing endpoints:**
- `POST /users` accepts an `organizations` field to create memberships/role bindings at creation time.
- `POST /sessions/validate`, `GET /users/{id}`, and `GET /users` now include the organizations a user belongs to and the roles held in each.
- Deleting a role cascades to its role bindings across every organization it was held in, matching the existing cascade behavior for organization and user deletion.

**Tradeoffs / things worth a close look:**
- `GET /users/{id}` and `GET /users` compute organizations via real pop associations (`OrganizationMemberships`, `RoleBindings` on `models.User`, populated by a new `AfterEagerFind` hook), rather than a hand-written SQL join — chosen for consistency with how every other association on `models.User` already works, at the cost of a few more queries per request. A working raw-SQL version exists as its own, earlier commit and is a one-command revert if the cost turns out to matter in practice.

# Tests

- Persister- and handler-level tests for every new endpoint, including cascade behavior, id-or-slug role resolution, and uniqueness constraints.
- Dedicated cross-tenant isolation coverage, run through multi-tenant routing since it requires two reachable tenants: organization/role name uniqueness scoped per tenant, cross-tenant 404s, and rejecting cross-tenant references when binding roles or adding members.


backend/handler/organization_public.go
```diff
@@ -0,0 +1,80 @@
...
+	var body dto.RoleCheckRequest
+	if err := (&echo.DefaultBinder{}).BindBody(c, &body); err != nil {
+		return dto.ToHttpError(err)
+	}
+
+	if err := c.Validate(body); err != nil {
+		return dto.ToHttpError(err)
+	}
+
+	membership, err := h.persister.GetOrganizationMembershipPersister().Get(userId, body.OrganizationID, tenant.ID)
```
FreddyDevelop, line 60: The `userId` is the `public_id` whereas `Get()` is expecting the internal id.

backend/handler/session.go
```diff
@@ -191,11 +198,55 @@ func (h *SessionHandler) ValidateSessionFromBody(c echo.Context) error {
...
 		UserID:         &claims.Subject,
 		IdleExpiresAt:  idleExpiresAt,
+		Organizations:  organizations,
 	})
 }
+
+// getOrganizations returns userID's organizations and role slugs,
+// computed fresh from the database on every call so a role or membership
+// change takes effect immediately, without reissuing the session token.
+func (h *SessionHandler) getOrganizations(userID uuid.UUID, tenantID uuid.UUID) ([]dto.ValidateSessionOrganization, error) {
```
FreddyDevelop: Also here, the callers of this method are using the public_id from the JWT, but the persister methods are expecting the internal id.

backend/handler/organization_admin.go
```diff
@@ -0,0 +1,402 @@
...
+
+	organization, err := h.persister.GetOrganizationPersister().Get(organizationId, tenant.ID)
+	if err != nil {
+		return fmt.Errorf("failed to get organization: %w", err)
+	}
+	if organization == nil {
+		return echo.NewHTTPError(http.StatusNotFound, "organization not found")
+	}
+
+	user, err := h.persister.GetUserPersister().Get(userId, tenant.ID)
```
FreddyDevelop: Here the `:user_id` path segment is interpreted as the internal id, but all other public facing endpoints expect the `public_id.

backend/handler/role_binding_admin.go
```diff
@@ -0,0 +1,183 @@
...
+				return echo.NewHTTPError(http.StatusConflict, "user already holds this role in this organization")
+			}
+		}
+		return fmt.Errorf("failed to create role binding: %w", err)
+	}
+
+	return c.JSON(http.StatusOK, admin.FromRoleBindingModel(binding, *role))
+}
+
+func (h *RoleBindingHandlerAdmin) List(c echo.Context) error {
```
FreddyDevelop, line 114: When org_id and/or user_id do not exist, the endpoint would return an empty list. But all other endpoints, e.g. `Create` and `Delete` in this file would return `404 - Not found`. It would be more consistent if this also returns `404 - Not found`.

backend/persistence/role_binding_persister.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch role bindings: %w", err)
+	}
+
+	return bindings, nil
+}
+
+// ListSlugsByUserGroupedByOrganization joins role_bindings to roles once,
+// filtered by user and tenant only (no organization_id filter), then
+func (p *roleBindingPersister) ListByUser(userID uuid.UUID, tenantID uuid.UUID) ([]models.RoleBinding, error) {
```
FreddyDevelop: The doc comment names the wrong method and is cut off mid-sentence.

FreddyDevelop (commented):
I think this PR does not include the new user.public_id field properly.
I flagged some places, but that might not be all.

backend/handler/organization_admin.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	var body admin.CreateOrganization
+	if err := (&echo.DefaultBinder{}).BindBody(c, &body); err != nil {
+		return dto.ToHttpError(err)
+	}
+
+	if err := c.Validate(body); err != nil {
+		return dto.ToHttpError(err)
+	}
```
FreddyDevelop: The `Patch` endpoint prevents users from changing the organizations name to just whitespaces. We should also prevent creating organizations with just whitespaces as a name.

backend/handler/role_admin.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	var body admin.CreateRole
+	if err := (&echo.DefaultBinder{}).BindBody(c, &body); err != nil {
+		return dto.ToHttpError(err)
+	}
+
+	if err := c.Validate(body); err != nil {
+		return dto.ToHttpError(err)
+	}
```
FreddyDevelop: The `Patch` endpoint prevents changing the name to only whitespaces. So we should prevent creating roles with only whitespaces as name.

backend/handler/role_binding_admin.go
```diff
@@ -0,0 +1,215 @@
...
+			if pgErr.Code == "23505" {
+				return echo.NewHTTPError(http.StatusConflict, "user already holds this role in this organization")
+			}
+		} else if mysqlErr, ok2 := errors.AsType[*mysql.MySQLError](err); ok2 {
+			if mysqlErr.Number == 1062 {
+				return echo.NewHTTPError(http.StatusConflict, "user already holds this role in this organization")
+			}
+		}
+		return fmt.Errorf("failed to create role binding: %w", err)
+	}
```
FreddyDevelop: The pg 23505/mysql 1062 unique-violation-to-409 detection block is duplicated near-verbatim across 4 new Create/Update handlers in this PR, growing the codebase-wide copy count to 6. 
We should make a util or helper function so we don't duplicate this more and more.

backend/handler/role_admin.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// A slug that parses as a UUID would be permanently unreachable via
+	// RolePersister.GetByIDOrSlug, which always tries UUID-parsing first.
+	if _, err := uuid.FromString(body.Slug); err == nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "slug must not be a valid uuid")
+	}
```
FreddyDevelop: A slug should also not include specific characters e.g. whitespaces, because the slug can be used in the path of a url (e.g. in removing a role binding).

We could use a regex pattern for this, e.g. `^[a-z0-9]+(?:-[a-z0-9]+)*$`.

backend/persistence/organization_persister.go
```diff
@@ -0,0 +1,118 @@
...
+	db *pop.Connection
+}
+
+func NewOrganizationPersister(db *pop.Connection) OrganizationPersister {
+	return &organizationPersister{db: db}
+}
+
+func (p *organizationPersister) Get(id uuid.UUID, tenantID uuid.UUID) (*models.Organization, error) {
+	organization := models.Organization{}
+	err := p.db.Where("organizations.tenant_id = ?", tenantID).Find(&organization, id)
```
FreddyDevelop: I'm not sure if it is intentional, but why `organizations.tenant_id` and not just `tenant_id` as in the other methods of this persister?

backend/persistence/role_binding_persister_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	membershipB, err := s.Storage.GetOrganizationMembershipPersister().Get(user1bID, org1bID, tenant1ID)
+	s.Require().NoError(err)
+	s.NotNil(membershipB)
+}
+
+// This is the FK that makes "removing a user from an organization cascades to
+// that user's role_bindings for that organization" actually true at the DB
+// level - the least obvious FK in the whole design, and the fastest possible
+// signal if it's ever missing or dropped in a future migration.
```
FreddyDevelop: This comment does not seem like a test description.

backend/persistence/role_persister.go
```diff
@@ -0,0 +1,129 @@
...
+	db *pop.Connection
+}
+
+func NewRolePersister(db *pop.Connection) RolePersister {
+	return &rolePersister{db: db}
+}
+
+func (p *rolePersister) Get(id uuid.UUID, tenantID uuid.UUID) (*models.Role, error) {
+	role := models.Role{}
+	err := p.db.Where("roles.tenant_id = ?", tenantID).Find(&role, id)
```
FreddyDevelop: Here also, why `roles.tenant_id` and not just `tenant_id`?

PR #2916, opened 2026-09-13
fix(backend): truncate User-Agent to 255 characters to prevent DB insert failure (#2877)

### LINKED ISSUE #2877: Login and registration return 500 when the User-Agent exceeds 255 characters

Tyagiquamar:
Fixes #2877


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

lfleischmann:
Thanks for the contribution! Closing in favour of already merged https://github.com/teamhanko/hanko/pull/2882.

PR #2918, opened 2026-09-14
fix(redis): apply the database index from the redis address

rluisr:
# Description

`RedisConfig.Address` is documented as taking an optional database index:

https://github.com/teamhanko/hanko/blob/main/backend/config/config_shared.go#L8

```go
// `address` is the address of the redis instance in the form of `host[:port][/database]`.
Address string `yaml:"address" json:"address" koanf:"address"`
```

The suffix is never applied. Both consumers pass `Address` straight to
`redis.Dial`, which does not parse it — only `redis.DialURL` does — so no
`SELECT` is ever issued and everything lands in database 0.

It does not fail as a silent fallback, which would be the benign outcome. The
whole string is used as the dial target, so the index is parsed as part of the
port:

```
rate_limiter:
  enabled: true
  store: redis
  redis_config:
    address: redis:6379/9
```

```
could not acquire lock: 1 error occurred:
	* node #0: dial tcp: lookup tcp/6379/9: unknown port
```

Every affected request then returns 500 `technical_error`. redigo dials
lazily, so the process starts up and reports healthy; the failure only appears
when the first rate-limited or flow-locked request arrives, which can be well
after a deployment looks green.

`redis://redis:6379/9` does not work either (`too many colons in address`), so
there is currently no way to place Hanko in a non-zero database.

This matters when Redis is shared with other services, where a dedicated
logical database is the usual way to keep key spaces separate and make
`FLUSHDB` safe.

# Implementation

Added `RedisConfig.DialAddress()`, which splits the configured address into
the part `redis.Dial` accepts and the database index, and applied the index
via `redis.DialDatabase` at both call sites:

- `rate_limiter.NewRateLimiter`
- `flow_locker.NewFlowLocker` → `NewRedisLocker` (new `Database` field on
  `RedisLockerConfig`)

Notes on the parsing:

- `strings.LastIndex` for the separator, so IPv6 literals such as
  `[::1]:6379/4` split correctly.
- A trailing slash (`localhost:6379/`) is treated as unspecified, i.e.
  database 0, matching `redis.DialURL`.
- A non-numeric or negative index is rejected with a clear config error rather
  than being passed through to fail later as an opaque dial error.
- Addresses with no suffix take an early return and are dialled exactly as
  before, so existing configurations are unaffected.

The flow locker returns the error (`NewFlowLocker` already returns one); the
rate limiter uses `log.Fatal`, consistent with how it already handles store
construction failures in that function.

# Tests

Parser unit tests in `backend/config/config_shared_test.go` cover host-only,
host with database, host without port, explicit zero, trailing slash, IPv6
with and without a database, and the non-numeric and negative error cases:

```
go test ./config/... ./rate_limiter/...
ok  	github.com/teamhanko/hanko/backend/v3/config	0.041s
ok  	github.com/teamhanko/hanko/backend/v3/rate_limiter	0.024s
```

I also drove the real constructors against a live Redis with
`address: 127.0.0.1:6380/9`, taking a token from the rate limiter and
acquiring a flow lock:

```
db0 keys: 0
db9 keys: 2
OK: all keys in db9, db0 untouched
```

The same check against unpatched `main` fails at the dial, which is the
behaviour described above:

```
lock error: node #0: dial tcp: lookup tcp/6380/9: unknown port
```

# Todos

None.

# Additional context

Found while moving a Hanko deployment to two instances, where `rate_limiter`
and `flow_locker` have to move off their `in_memory` defaults to be correct
across processes. With `in_memory`, each instance keeps its own counters — in
my testing the 6th wrong password was blocked on one instance while the first
attempt for the same user and IP still went through on the other.

An alternative implementation would be to route through `redis.DialURL` and
accept full `redis://` URLs, which would also bring TLS via `rediss://`. I
kept this change narrow so the documented format starts working without
altering what `address` accepts.


backend/config/config_shared.go
```diff
@@ -11,6 +17,37 @@ type RedisConfig struct {
 	Password string `yaml:"password" json:"password,omitempty" koanf:"password"`
 }
 
+// DialAddress splits Address into the `host[:port]` that can be passed to
+// redis.Dial and the optional database index from the `/database` suffix.
```
lfleischmann, line 21: First off: nice catch  — could reproduce (error and fix)                                                                                                                
                                                                                                                                                     
One question though, since you already mention it here in the comment: if `redis.DialURL` already does this parsing, why not use                                                                                                         
                                                                                                                                                     
```go                                                                                                                                              
  conn, err := redis.DialURL("redis://"+cfg.Redis.Address, redis.DialPassword(cfg.Redis.Password)) 
```                                                  
                                                                                                                                                     
then instead of the hand-rolled `DialAddress()` parser?  I tested it against the same cases (host:port, host:port/db, trailing slash, IPv6, non-numeric/negative db) and it matches this PR's behavior — while also handling a case this PR doesn't: a host with no port (valid per the  documented host[:port][/database] format, e.g. redis/3). DialAddress() splits that correctly but then hands the bare host to redis.Dial, which fails with missing port in address; `DialURL` defaults the missing port to 6379 on its own.

